Issue #11, opened 2018-04-22
Add a 'copy' deployment mode for cargo-screeps


jechase:
I often find myself wanting to deploy to a local sandbox server that I'm running, and sometimes can't be bothered to set up the authentication mod that allows for the http upload functionality to work.

It would be nice to have an option to instead have the build artifacts copied to the directory that the screeps client watches and uploads for you.

The config could be something like:

```toml
mode = "copy" # Default: copy, valid: copy|upload
branch = "default"

[upload]
username = "user"
password = "pass"
# optional:
# ptr = false
# hostname = "localhost"
# ssl = true
# port = 443

[copy]
dest = "/path/to/screeps/client/scripts" # path to directory containing the branch 
                                         # directories, configured branch should be
                                         # appended
prune = true # whether or not to clear the destination of extra files
```

Edit: This is probably something that I'd be willing to work on myself.

daboross:
Sounds like a good plan!

I think this could work well with also adding the ability to use multiple configuration files.

Assuming we add this, `[upload]` should be completely optional when `mode = "copy"` and `[copy]` is completely optional when we add `mode = "upload"`?

I think we could rename the `upload` option to `deploy`, then, and just do the configured mode.

PR #12, opened 2018-04-22
Configurable deployment modes

jechase:
This implements #11 

### Configuration changes

* A `mode` key is added at the top level. It can be either "copy" or "upload", with "upload" being the default for backwards-compat
* `[copy]` and `[upload]` sections. 
  * `[copy]` has the required `dest` and an optional `prune` key.
    * `dest` should refer to the directory *containing* the branch directories, and the top-level `branch` key will determine which of these to use.
  * `[upload]` has all of the original top-level keys except for `branch`, `output_wasm_file`, and `output_js_file`
* All of the `[upload]` section's keys are still valid for the top level for backwards-compat, but will be ignored if the section is present.

### Functionality changes

Old function flags are now subcommands.

* `deploy`: Run whichever deployment mode is configured
* `upload`: Changed to run in upload mode regardless of configured mode
* `copy`: Run in copy mode regardless of configured mode


jechase:
OK, I've got this just about to where I wanted to go with it. There's bound to be something that I've missed, so lmk what you think.

I'll also clean up the commit history once it's all approved and before merging.

cargo-screeps/src/main.rs
```diff
@@ -52,7 +54,7 @@ fn run() -> Result<(), failure::Error> {
 
 fn main() {
     if let Err(e) = run() {
-        eprintln!("{}", e.backtrace());
+        // eprintln!("{}", e.backtrace());
```
daboross: If we want to remove this, probably best to either directly remove it?

cargo-screeps/src/orientation.rs
```diff
@@ -3,21 +3,16 @@ use std::{env, path::PathBuf};
...
 
 pub fn find_project_root() -> Result<PathBuf, failure::Error> {
-    // TODO: is it right to canonicalize here?
-    let mut current = env::current_dir()?.canonicalize()?;
+    let orig = env::current_dir()?;
+    let mut current = orig.clone();
 
     loop {
-        if current.join("screeps.toml").exists() {
+        if current.join("Cargo.toml").exists() {
```
daboross: One reason I was checking for screeps.toml here originally is in order to support correctly building if we're running in a sub-directory which is also a crate (most likely a library used by the top-level deployed crate).

cargo-screeps/src/setup.rs
```diff
@@ -1,56 +1,70 @@
...
+                        } else {
+                            cfg_arg
+                        }
+                    })
+                    .subcommand(
+                        clap::SubCommand::with_name("build")
+                            .about("build files, put in target/ in project root"),
+                    )
+                    .subcommand(
+                        clap::SubCommand::with_name("check")
```
daboross: Thanks for fixing this up! It was annoying to have it as "--check", etc.

cargo-screeps/src/orientation.rs
```diff
@@ -3,21 +3,16 @@ use std::{env, path::PathBuf};
...
-        let is_last = current.pop();
+        let is_last = !current.pop();
         if is_last {
-            bail!(
-                "could not find 'screeps.toml' in {} or parents.\n\
-                 Please create 'screeps.toml' in project root. (template at \
-                 https://github.com/daboross/screeps-in-rust-via-wasm/blob/master/screeps.toml)",
-                current.display()
-            );
+            bail!("could not find crate root {} or parents", orig.display());
```
daboross: Ideally I'd like to still point people to the default configuration file when they don't have one. Even if where we actually check for the configuration moves around, the error message for when we don't have a screeps.toml should probably still include this?

cargo-screeps/src/config.rs
```diff
@@ -0,0 +1,178 @@
...
+            config_file.exists(),
+            "expected configuration to exist at {}",
+            config_file.display(),
+        );
+
+        let config_str = {
+            use std::io::Read;
+            let mut buf = String::new();
+            fs::File::open(config_file)
+                .context("opening config file")?
```
daboross: Oh right - one other thing. Failure's `.context()` is not going to be too useful, since the new error message does not actually include the original one - it just replaces whatever the error was with "opening config file" without any mention of the OS error.

I'd prefer to use specific error variants instead if that's alright. We could also change the handler in `main` to print out all error causes, that would also work.

daboross:
It's been a while, so I'll merge this and just make some tweaks afterwards to fix the points I mentioned in the above review. Hope that's OK.

These are good changes, and I'm glad to have them. Thanks!

ASalvail:
@daboross what's your intention for this PR? It'd be quite nice to have it merged.

daboross:
Darn it, after doing a more thorough review I think there are more changes I'd like to make to this than not.

Having the separate configuration structures and abusing `#[serde(default)]` is definitely irking me here. I'd much rather break backwards compatbility than have tons of extra code to handle odd cases.

Also, pulling in file configuration into `setup_cli!()` really breaks separation of concerns. I'd rather have that as purely cli stuff- not file configuration.

I'm glad to have this as implemented as much as it is, but as of right now I'm going to be making some major changes before pulling it in. I'll open up a new PR superceding this one when it's ready.

PR #15, opened 2018-09-05
Added structure_controller module

ASalvail:
I tried my hand at extending the current object implementations with a `structure_controller`.
I inspired myself heavily on other modules and did test the methods, but a good look is warranted.

### DIFF
--- a/screeps-game-api/src/objects/impls/structure_controller.rs
+++ b/screeps-game-api/src/objects/impls/structure_controller.rs
@@ -1,0 +1,58 @@
+use stdweb::Value;
+
+use objects::StructureController;
+
+simple_accessors! {
+    StructureController;
+    (level -> level -> u32),
+    (progress -> progress -> u32),
+    (progress_total -> progressTotal -> u32),
+    (safe_mode -> sameMode -> u32),
+    (safe_mode_available -> sameModeAvailable -> u32),
+    (safe_mode_cooldown -> sameModeCooldown -> u32),
+    (ticks_to_downgrade -> ticksToDowngrade -> u32),
+    (upgrade_blocked -> upgradeBlocked -> u32)
+}
+
+#[derive(Debug)]
+pub struct Reservation {
+    pub username: String,
+    pub ticks_to_end: u32
+}
+
+#[derive(Debug)]
+pub struct Sign {
+    pub username: String,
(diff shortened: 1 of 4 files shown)


screeps-game-api/src/objects/impls/structure_controller.rs
```diff
@@ -0,0 +1,58 @@
...
+pub struct Sign {
+    pub username: String,
+    pub text: String,
+    pub time: u32,
+    pub datetime: String  // todo: use real date type
+}
+
+impl StructureController {
+    pub fn reservation(&self) -> Option<Reservation> {
+        if let Value::Reference(r) = js!(return @{self.as_ref()}.reservation;){
```
daboross: For style I'd use a `match` with a catch-all `_` branch rather than `if let .. else`, but that's not a big deal.

daboross:
Thank you for making this!

daboross:
Sorry, I seemed to have forgotten a design decision when pulling this in. `prelude` is meant purely for traits needed to use methods on structures, not the data structures themselves, so I'm going to remove Reservation and Sign from there. It's meant as a bare minimum number of imports for functioning- for using types like StructureController, Creep, Sign, Reservation, etc. importing from crate root is required instead. Hope that's alright.

ASalvail:
For sure. As this is... 🤔 lightly documented, I go from what I can deduce. Thanks for pointing that out, that'll help for the next one (probably Lab... or Market). I can push the changes tonight even, if you'd like.

screeps-game-api/src/objects/impls/structure_controller.rs
ASalvail: Any reasons why? Still learning my way around Rust and that line was harder to write and debug than I'd care to admit.

The thing I could see is being able to extend match going forward.

daboross:
"lightly documented" is putting it lightly! This thing was definitely thrown together without much thought on letting anyone know what it's doing- I wanted to get something out here, and work on refining it later.

I'm glad it's not completely unreadable, though, since you got this written! I'd be very open to more changes, and if you have any questions about how/why things are done, feel free to ask. I _think_ I have a reason for most things in here, even if that reason is "this is the fastest way I could write this code, so feel free to fix it up".

PR #17, opened 2018-09-07
Implements raw_memory getter and setter.

ASalvail:
Pretty much self explanatory.


daboross:
Thanks!

I *think* it'd be safe to have `get` return a `String` rather than `Option<String>`. For the segments, they may be unset / null, but as I understand it `RawMemory.get()` will always return a string, even if it's an empty one. This would just be changing `-> Option<String>` to `String`, if you'd want to make the change.

ASalvail:
Done. Next are ids...

ASalvail:
I build against my local repository, so I'm not at all affected.

PR #18, opened 2018-09-07
Support for PartialEq and Eq for objects with an id

ASalvail:
This PR adds methods to fetch `id`s on the object where this was missing.

Given those, it also implements `PartialEq` and `Eq` for all objects with an id.
Finally, it also adds `PartialEq` and `Eq` for `Room` based on name.

All of this for partitioning creeps by rooms...

### DIFF
--- a/screeps-game-api/src/objects/impls/creep.rs
+++ b/screeps-game-api/src/objects/impls/creep.rs
@@ -24,6 +24,10 @@ impl Creep {
         js_unwrap!(@{self.as_ref()}.carry[RESOURCE_ENERGY])
     }
 
+    pub fn owner_name(&self) -> String {
+        js_unwrap!(@{self.as_ref()}.owner.username)
+    }
+
     pub fn cancel_order(&self, name: &str) -> ReturnCode {
         js_unwrap!(@{self.as_ref()}.cancelOrder(@{name}))
     }
@@ -148,6 +152,7 @@ macro_rules! creep_simple_concrete_action {
 
 simple_accessors! {
     Creep;
+    (id -> id -> String),
     (carry_capacity -> carryCapacity -> i32),
     (fatigue -> fatigue -> i32),
     (hits -> hits -> i32),
@@ -157,6 +162,15 @@ simple_accessors! {
     (spawning -> spawning -> bool),
     (ticks_to_live -> ticksToLive -> i32),
 }
+
+impl PartialEq for Creep {
+    #[inline]
+    fn eq(&self, other: &Creep) -> bool{
+        self.id() == other.id()
+    }
+}
+
+impl Eq for Creep {}
 
 creep_simple_generic_action! {
     (attack(Attackable) -> attack),
--- a/screeps-game-api/src/objects/impls/creep.rs
+++ b/screeps-game-api/src/objects/impls/creep.rs
@@ -164,7 +164,6 @@ simple_accessors! {
 }
 
 impl PartialEq for Creep {
-    #[inline]
     fn eq(&self, other: &Creep) -> bool{
         self.id() == other.id()
     }
--- a/screeps-game-api/src/objects/macros.rs
+++ b/screeps-game-api/src/objects/macros.rs
@@ -88,21 +88,23 @@ macro_rules! impl_room_object_properties {
 }
 
 
-// Macro for mass implementing `StructureProperties`, `PartialEq` and `Eq` for a type. 
-// 
-// This macro accepts a comma-separated list of types on which to implement the unsafe `StructureProperties` trait on
-// a screeps object. 
-// From that implementation, the type gets the `id` method which is used to implement `PartialEq` and `Eq`.
-// 
-// # Example
-// ```
-// macro_rules!(OwnedStructure, Structure, StructureContainer)
-// ```
-// 
-// # Safety
-// The macro assumes that it is implementing the trait to a valid `Reference` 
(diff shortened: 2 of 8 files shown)


screeps-game-api/src/objects/impls/creep.rs
```diff
@@ -24,6 +24,10 @@ impl Creep {
         js_unwrap!(@{self.as_ref()}.carry[RESOURCE_ENERGY])
     }
 
+    pub fn owner_name(&self) -> String {
```
daboross: It'd be nice to match this naming with `OwnedStructureProperties::owner` and `ConstructionSite::owner`. I think I might just change those two, though, since `owner_name` is strictly more accurate.

screeps-game-api/src/objects/macros.rs
```diff
@@ -86,3 +86,33 @@ macro_rules! impl_room_object_properties {
         )*
     );
 }
+
+
+// Macro for mass implementing `StructureProperties`, `PartialEq` and `Eq` for a type.
```
daboross: I think this could be a doc comment with `//!`, but it's not a big deal - I can add in the exclamation marks afterwards.

screeps-game-api/src/objects/impls/creep.rs
```diff
@@ -158,6 +163,14 @@ simple_accessors! {
     (ticks_to_live -> ticksToLive -> i32),
 }
 
+impl PartialEq for Creep {
+    fn eq(&self, other: &Creep) -> bool{
```
daboross: Hope it's alright if I run `cargo fmt` on this afterwards- it'd just add spacing here between return types and the opening brace for the method.

screeps-game-api/src/objects/macros.rs
```diff
@@ -86,3 +86,33 @@ macro_rules! impl_room_object_properties {
...
+
+// Macro for mass implementing `StructureProperties`, `PartialEq` and `Eq` for a type. 
+// 
+// This macro accepts a comma-separated list of types on which to implement the unsafe `StructureProperties` trait on
+// a screeps object. 
+// From that implementation, the type gets the `id` method which is used to implement `PartialEq` and `Eq`.
+// 
+// # Example
+// ```
+// macro_rules!(OwnedStructure, Structure, StructureContainer)
```
daboross: Unless I'm missing something, I think this line was meant to be something else?

daboross:
Glad to have you improving this repo! Hope my PR reviews aren't too scrutinous.

screeps-game-api/src/objects/macros.rs
ASalvail: I changed the `//` for `///`, but I'll wait until #19 is merged before moving this one with the other.

screeps-game-api/src/objects/impls/creep.rs
ASalvail: I tried that once. It changed so much that I reverted it. I'll endeavor to run it next time prior to making a PR...

screeps-game-api/src/objects/impls/creep.rs
```diff
@@ -148,6 +152,7 @@ macro_rules! creep_simple_concrete_action {
 
 simple_accessors! {
     Creep;
+    (id -> id -> String),
```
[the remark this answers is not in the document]
daboross: That's a good idea!

I've been using github issues for this in other projects, they seem to work well after the issue labels have been massaged a bit. I'll open one for `HasId`, and we can go from there.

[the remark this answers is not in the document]
daboross: Opened #23.

screeps-game-api/src/objects/macros.rs
ASalvail: The copy/paste monster is real!

ASalvail:
Absolutely not, it's all very constructive. Still learning the ropes of rust (which I really like), so your comments help me getting a feel for what is usually left unsaid and best practices.

Thanks for taking the time to review my small avalanche of PRs!

screeps-game-api/src/objects/impls/creep.rs
daboross: Ah yeah, I guess its style has probably changed quite a bit since I last ran it on this repository.

I'll run and commit the change myself. I'll do it after merging this batch of PRs, though, so that the formatting changes don't make them unmergable.

screeps-game-api/src/objects/impls/creep.rs
daboross: It might be nice, but I'm not too concerned with typing everything that's a string value underneath.

My main concern is that while things like `Username` or `ObjectId` could be useful within one instance, any screeps AI will need to serialize them and retrieve them from memory every tick, and at that point there's no more safety than just treating them as strings anyways.

I've opened https://github.com/daboross/screeps-in-rust-via-wasm/issues/29.

daboross:
Glad to hear it's constructive!

Reviewing a small avalanche of PRs is only a tiny bit of work compared to the benefit of extra eyes on the project and the contributions themselves. Thanks for coming and improve `screeps-game-api`!

daboross:
I'll merge this manually since the changes in `src/objects/macros.rs` conflict with the macro cleanup.

ASalvail:
Thanks! I was going through the changes. I'll leave it to you. 
We should coordinate on slack next time 🤔

daboross:
Ah yeah- that could work!

PR #19, opened 2018-09-08
Macro documentation

ASalvail:
My biggest hurdle when starting to use this library (aside from the doc not building on [doc.rs](https://docs.rs/crate/screeps-game-api/) was reading through the exhaustive use of macros.

Being fairly new to rust, I had to learn the somewhat esoteric macro grammar (which in hindsight is fairly simple) in order to be able to read the macro, then do mental substitution in order to figure out what the macros were used for.

I hope those modest comments are going to help the next person reading through the source code.

If I may suggest, I'd move all macros in a `macro.rs` file living at the root of the package and `use`d by `lib.rs`. Since I learned that the order in which the files are parsed matters for the macro, I can just imagine clueless me try and figure out on the fly why a macro is out of scope. Also, I think it'd clean the codebase a bit.


screeps-game-api/src/lib.rs
```diff
@@ -11,13 +11,71 @@ extern crate serde_derive;
...
+Here, `js_unwrap!()` takes any valid javascript expression (expresses a value) 
+and will attempt conversion to the receiving variable type using `try_into`. 
+For example:
+```
+let s: u32 = js_unwrap!(Game.time);
+```
+
+This will be be converted to 
+```
+let s: u32 = js!(Game.time).try_into().expect('Some Err Msg');
```
daboross: I think this should be `js!(return Game.time)` rather than `js!(Game.time)`. The latter will not actually work, since `return` is necessary to return any values.

That was part of the motivation for having `js_unwrap!` - not having write `return ` everywhere.

screeps-game-api/src/lib.rs
```diff
@@ -11,13 +11,71 @@ extern crate serde_derive;
 #[macro_use]
 extern crate stdweb;
 
+/*  Used to get data from a javascript reference back into rust code.
```
daboross: Style wise, I'd prefer doc comments with `//!` rather than than `/*`. Not a big deal, I can definitely change this after merging if you're alright with that.

The practical advantages of doc comments include code examples being tested with `cargo test` and inclusion in editor tooltips when using `rls` or other rust-IDE integration.

screeps-game-api/src/lib.rs
```diff
@@ -11,13 +11,71 @@ extern crate serde_derive;
...
+Since `Game.time` returns a javascript `number`, `js!` spits out a
+`stdweb::Value::Number` which is convertible to a u32 and should work without
+problem.
+
+A non-exhaustive list of types that work (use your judgement)
+
+  js      |  rust
+------------------------
+Number    | u32, i32, f32
+String    | String, &str
```
daboross: I don't think `&str` is usable as a result of `js_unwrap!()` since it's a borrowed value and rust has to own the memory it is borrowed from.

screeps-game-api/src/lib.rs
```diff
@@ -11,13 +11,71 @@ extern crate serde_derive;
...
+This will be be converted to 
+```
+let s: u32 = js!(Game.time).try_into().expect('Some Err Msg');
+```
+
+Since `Game.time` returns a javascript `number`, `js!` spits out a
+`stdweb::Value::Number` which is convertible to a u32 and should work without
+problem.
+
+A non-exhaustive list of types that work (use your judgement)
```
daboross: It might be worth linking to https://docs.rs/stdweb/0.4.8/stdweb/unstable/trait.TryFrom.html for a full list, if you want.

screeps-game-api/src/objects/macros.rs
```diff
@@ -37,52 +64,31 @@ macro_rules! simple_accessors {
                 }
             )*
         }
-    );
-    (
```
daboross: Good catch on this. I think these must have been from before all the traits had default implementations of getter methods.

daboross:
Thank you for this! These macros were made quite haphazardly.

Moving all of these into a top level `macros` module is a good idea too - it would definitely make it cleaner. This could be part of this PR, or another one, or I could add it in afterwards.

daboross:
Speaking of lack of `docs.rs`, that is an annoyance, but if you need it it's pretty easy to build documentation locally too.

Try running `cargo doc --open` in the `screeps-starter-rust` repository, or your own project repository. It should generate documentation for the crates being used and open it up, and that should include documentation for `screeps-game-api`.

I guess this tip belongs in the `README` now that docs.rs is failing.

screeps-game-api/src/lib.rs
daboross: Darn it, meant `///` here instead of `//!`. Not sure what I was thinking.

ASalvail:
Thanks for the feedback. I addressed most comments and moved the macros to a single file which I believe will make the code clearer.

> These macros were made quite haphazardly.
I'd have taken the easy way out and copy pasted the code, so it served as a good incentive to learn about macros. I added in the doc the most helpful link I found on macros.

The doc compiles, but I believe it skips over the macros since they are not exported. Still, it'll be helpful for the next contributor.

screeps-game-api/src/game.rs
```diff
@@ -202,73 +202,34 @@ pub mod shard {
...
-        )*
-    };
-}
-
 game_map_access! {
     /// See [http://docs.screeps.com/api/#Game.constructionSites]
     ///
     /// [http://docs.screeps.com/api/#Game.constructionSites]: http://docs.screeps.com/api/#Game.constructionSites
-    (construction_sites, objects::ConstructionSite, Game.constructionSites),
+    construction_sites, objects::ConstructionSite, Game.constructionSites;
```
daboross: I think we should keep macro syntax changes separate from the reorganization, at least for good history.

If we want to do some normalization on macro usage syntax, that would be a good idea, but I think it would be more useful to do and discuss in a new PR.

daboross:
Cool, thanks! LGTM besides that syntax change in `game_map_access!`. Not necessarily bad, just that it would work better as a separate PR for discussion purposes.

ASalvail:
That's fair.
Here's the final version.

daboross:
Thanks!

PR #21, opened 2018-09-08
findClosestByRange returns an object, not an array

ASalvail:
Unless I am mistaken from the doc, `findClosestByRange()` returns an object, not an array of objects.

https://docs.screeps.com/api/#RoomPosition.findClosestByRange

### DIFF
--- a/screeps-game-api/src/constants.rs
+++ b/screeps-game-api/src/constants.rs
@@ -55,7 +55,7 @@ impl TryFrom<Value> for ReturnCode {
 }
 
 pub unsafe trait FindConstant {
-    type Item: TryFrom<Value, Error = <Reference as TryFrom<Value>>::Error>;
+    type Item: TryFrom<Value, Error = <Reference as TryFrom<Value>>::Error> + AsRef<Reference>;
 
     fn find_code(&self) -> i32;
 }
--- a/screeps-game-api/src/objects/impls/room_position.rs
+++ b/screeps-game-api/src/objects/impls/room_position.rs
@@ -54,11 +54,11 @@ impl RoomPosition {
             .expect("expected Flag.createFlag to return ReturnCode")
     }
 
-    pub fn find_closest_by_range<T>(&self, ty: T) -> Vec<T::Item>
+    pub fn find_closest_by_range<T>(&self, ty: T) -> T::Item
     where
         T: FindConstant,
     {
-        js_unwrap_array!(@{self.as_ref()}.findClosestByRange(
+        js_unwrap!(@{self.as_ref()}.findClosestByRange(
             __structure_type_num_to_str(@{ty.find_code()}
         )))
     }
--- a/screeps-game-api/src/objects/impls/room_position.rs
+++ b/screeps-game-api/src/objects/impls/room_position.rs
@@ -58,9 +58,9 @@ impl RoomPosition {
     where
         T: FindConstant,
     {
-        js!(@{self.as_ref()}.findClosestByRange(
-            __structure_type_num_to_str(@{ty.find_code()})
-            )).try_into().ok()
+        js_unwrap!(@{self.as_ref()}.findClosestByRange(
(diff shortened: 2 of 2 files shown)


daboross:
:+1: this is indeed a mistake.

While we're fixing it, though, I think it should return `Option<T::Item>` rather than `T::Item` since the method can fail and return `null`.

daboross:
One cool thing about `TryInto` is that it is already implemented for `Option<T>` ty es.

If you're alright with it, I'd rather we do use `js_unwrap!()`, even when the result is an `Option`, since that way a valid reference is translated to `Some(..)`, null or undefined is translated to `None`, and something unexpected like a string or an integer results in a panic. With `.try_into().ok()`, we incorrectly treat strings and integers as null references rather than panicking.

ASalvail:
This is what is the hardest still for me to grok: the From/Into magic.
You'll have to give me a hand for this one since:

```rust
    pub fn find_closest_by_range<T>(&self, ty: T) -> Option<T::Item>
    where
        T: FindConstant,
    {
        js_unwrap!(@{self.as_ref()}.findClosestByRange(
            __structure_type_num_to_str(@{ty.find_code()}))
        )
    }
```
results in 
```
error[E0277]: the trait bound `<T as constants::FindConstant>::Item: std::convert::AsRef<stdweb::Reference>` is not satisfied
  --> screeps-game-api\src\lib.rs:16:9
   |
16 |           ::stdweb::unstable::TryInto::try_into(js! { return $($code)* })
   |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::convert::AsRef<stdweb::Reference>` is not implemented for `<T as constants::FindConstant>::Item`
   |
  ::: screeps-game-api\src\objects\impls\room_position.rs:61:9
   |
61 | /         js_unwrap!(@{self.as_ref()}.findClosestByRange(
62 | |             __structure_type_num_to_str(@{ty.find_code()}))
63 | |         )
   | |_________- in this macro invocation
```

daboross:
Ohh, hm. I might have been assuming something untrue!

Let me look into this.

daboross:
I think the reason it's complaining about that is this implementation in `stweb`:

```rust
impl<T: TryFrom<Value, Error = ConversionError> + AsRef<Reference>> TryFrom<Value> for Option<T>
```

There are other implementations for Option wrapping primitive numbers and String, but this is the implementation we'd need to use to have it automatically unwrap our return values.

I had thought the generic implementation was for anything, but apparently it's only for things implementing `AsRef<Reference>`. I'm not 100% sure why they chose this, but I guess it's to limit automatic null checks to things which can be null.

Lucky for us, everything we want to return does implement `AsRef<Reference>`. The only problem is that this is a generic method, and the trait doesn't explicitly require the return type (`T::Item`) to be `AsRef<Reference>`.

On line 302 of constants.rs, `FindConstant` requires that the `Item` is `TryFrom<Value>`:

https://github.com/daboross/screeps-in-rust-via-wasm/blob/f43e2b5bf82c4305eb419f2f224f1a2045550f9d/screeps-game-api/src/constants.rs#L301-L305

Adding `+ AsRef<Reference>` to that requirement fixes the error:

```rust
pub unsafe trait FindConstant {
    type Item: TryFrom<Value, Error = <Reference as TryFrom<Value>>::Error> + AsRef<Reference>;

    fn find_code(&self) -> i32;
}
```

---

General From/Into magic is pretty magic, but it all boils down to trait implementations. I'm not sure I can understand all the interactions, but I've gotten at least a little good at tracking down what's causing errors.

If you want to make that change to `FindConstant`, you're welcome to do it. Otherwise I'll merge this as is and make the change to `FindConstant` and to using `js_unwrap!()` in a followup PR.

daboross:
No problem- it does take a while to get used to navigating all the little bits in Rust's type system.

PR #26, opened 2018-09-08
Add reasoning and documentation for unsafe traits

daboross:
Should address https://github.com/daboross/screeps-in-rust-via-wasm/issues/20.

Sending this as a PR for potential feedback and to have a CI check before merging.

### DIFF
--- a/screeps-game-api/src/objects/mod.rs
+++ b/screeps-game-api/src/objects/mod.rs
@@ -207,8 +207,8 @@ where
 ///
 /// # Contracts
 ///
-/// The reference returned by the `AsRef<Reference>` implementation must be a
-/// reference to a JavaScript object extending the `RoomObject` class.
+/// The reference returned by `AsRef<Reference>::as_ref` must reference a
+/// JavaScript object extending the `RoomObject` class.
 pub unsafe trait RoomObjectProperties:
     AsRef<Reference> + Into<Reference> + HasPosition
 {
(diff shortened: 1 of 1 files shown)


screeps-game-api/src/objects/mod.rs
```diff
@@ -175,6 +192,7 @@ impl HasPosition for RoomPosition {
     }
 }
 
+// All RoomObjects have positions.
```
ASalvail: `//` -> `///`
This has to be karma 😛

screeps-game-api/src/objects/mod.rs
```diff
@@ -194,6 +219,13 @@ pub unsafe trait RoomObjectProperties:
     }
 }
 
+/// Trait for all wrappers over Screeps JavaScript objects extending
+/// the `Structure` class.
+///
+/// # Contracts
```
ASalvail: Since the `unsafe` here is carried from `RoomObjectProperties`, is it worth repeating?

screeps-game-api/src/objects/mod.rs
daboross: :smile:

screeps-game-api/src/objects/mod.rs
daboross: Oh! I forgot to edit these to actually not just be copy-pasted versions of the one above, which they are.

Thanks for catching that - let me know what you think of repetitiveness if they're actually reasonably different.

ASalvail:
LGTM 👍

PR #27, opened 2018-09-08
Accept &str as an input to raw_memory::get

daboross:
Just an API niceness since we don't need to own the string. The value will need to be copied anyways when sending to JavaScript, so we might as well accept a borrowed string.

CC #17. I'd forgotten about this detail before.


ASalvail:
👍

PR #31, opened 2018-09-09
Macros for Memory use

daboross:
This is intended as a possible alternative to #22, possibly implemented in parallel with either option available for use.

This might also belong in a higher-level crate, though we can always implement it here then pull stuff out and reorganize later.

### DIFF
--- a/screeps-game-api/src/macros.rs
+++ b/screeps-game-api/src/macros.rs
@@ -450,9 +450,10 @@ macro_rules! mem_get {
 
 /// Set a value in memory given a path, creating dicts for intermediate places if they do not exist.
 ///
-/// # Panics
+/// # Return
 ///
-/// This macro will panic if any of the intermediate paths exist but are not dictionaries.
+/// This macro produces a `Result<(), ::screeps::memory::UnexpectedTypeError>`. The error path will
+/// trigger if any of the intermediate memory keys exist but are not dictionaries.
 ///
 /// # Examples
 ///
@@ -464,7 +465,7 @@ macro_rules! mem_get {
 ///
 /// # fn main() {
 /// let mem = screeps::memory::root();
-/// mem_set!(mem.creeps.John.count = 42);
+/// mem_set!(mem.creeps.John.count = 42).unwrap();
 /// # }
 /// ```
 ///
@@ -478,8 +479,8 @@ macro_rules! mem_get {
 /// let mem = screeps::memory::root();
 /// let creep_name = "John";
 /// let what_to_set = "count";
-/// mem_set!(mem.creeps[creep_name][what_to_set] = 51);
-/// mem_set!(mem.creeps[creep_name].count = 52);
+/// mem_set!(mem.creeps[creep_name][what_to_set] = 51).unwrap();
+/// mem_set!(mem.creeps[creep_name].count = 52).unwrap();
 /// # }
 /// ```
 #[macro_export]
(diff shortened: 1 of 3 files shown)


screeps-game-api/src/macros.rs
```diff
@@ -375,3 +375,182 @@ macro_rules! game_map_access {
...
+/// Accepted suffixes for type are methods that exist on `MemoryReference`, such as `num`, `int`,
+/// `string`, `bool`, `arr` and `dict`.
+#[macro_export]
+macro_rules! mem_get {
+    // Macro entry point
+    ($memory_reference:ident $($rest:tt)*) => {
+        mem_get!(@so_far { Some(&$memory_reference) } @rest $($rest)*)
+    };
+    // Access the last part with a variable
+    (@so_far { $reference_so_far:expr } @rest [ $next_part_variable:expr ] . $accessor:ident) => {
```
ASalvail: Might be clearer if you called the variable `$last_part_variable:expr`.

screeps-game-api/src/macros.rs
```diff
@@ -375,3 +375,182 @@ macro_rules! game_map_access {
...
+    ($($not_valid:tt)*) => {
+        compile_error!(concat!("Unexpected usage of mem_get! usage: ", stringify!($($not_valid)*)))
+    }
+}
+
+/// Set a value in memory given a path, creating dicts for intermediate places if they do not exist.
+///
+/// # Panics
+///
+/// This macro will panic if any of the intermediate paths exist but are not dictionaries.
```
ASalvail: Isn't that harsh? Maybe return an `Error` instead?

screeps-game-api/src/memory.rs
```diff
@@ -72,6 +72,29 @@ impl MemoryReference {
             .map(MemoryReference)
     }
 
+    /// Get a dictionary value or create it if it does not exist.
+    ///
+    /// If the value exists but is a different type, this will return `None`.
+    pub fn dict_or_create(&self, key: &str) -> Option<MemoryReference> {
+        (js!{
+            var map = (@{self.as_ref()})
+            var key = (@{key})
```
ASalvail: 2 missing `;`

screeps-game-api/src/macros.rs
daboross: I'd consider a path which should be a dictionary being an array or integer a hard error mostly because it should never happen outside of logic errors, and there isn't one generally good way to recover from it without the potential of loosing data.

If two parts of code are accessing the same memory path and expecting it to be different types, that's a bug. I mean we could return an Option or Err here, but I would expect that most usages of this utility wouldn't want to have to think about handing this situation besides just panicking themselves.

screeps-game-api/src/memory.rs
daboross: Right

screeps-game-api/src/macros.rs
ASalvail: I for one would rather it fails without panicking and write an `error!` to the logs.

Your argument basically boils down to: in order to not have corrupt memory, I'll stop executing my AI midway in the game loop. To me the latter is much worse, especially since I'm not given a choice (afaik, you can't catch a `panic`).

ASalvail: I mean, it *is* a bug, and should be fixed, but _in prod_, I'd rather a bit of lost memory than a non-functional colony.

daboross: True! I'll switch this to returning a `Result`.

daboross:
Thanks for the review for this @ASalvail!

I've changed the panic to a `Result<(), UnexpectedTypeError>` return.

If you're good with these changes, I'll merge this.

ASalvail:
Looks great! Thanks!

PR #33, opened 2018-09-09
Configurable deployment modes

### LINKED ISSUE #11: Add a 'copy' deployment mode for cargo-screeps

daboross:
This is a PR with my changes on top of @jechase's from #12 

Closes #12. Fixes #11.


cargo-screeps/src/copy.rs
```diff
@@ -0,0 +1,56 @@
...
+    let target_dir = root.join("target");
+
+    let mut deployed: HashSet<PathBuf> = HashSet::new();
+
+    for entry in fs::read_dir(target_dir)? {
+        let entry = entry?;
+        let path = entry.path();
+
+        if let (Some(name), Some(extension)) = (path.file_name(), path.extension()) {
+            if extension == "js" {
```
ASalvail: If at some point the user changes the configured file name, they'd end up with two "main" files and wasm files copied over.

Shouldn't that refer directly to the build config specific file names?

cargo-screeps/src/config.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub branch: String,
+    pub ssl: bool,
+    pub port: i32,
+    pub ptr: bool,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct CopyConfiguration {
+    pub destination: PathBuf,
+    pub branch: String,
```
ASalvail: Unless I'm mistaken, the `branch` isn't used (at least, not in `copy.rs`)

daboross: I think it should be, at least. It's used as a sub-directory under destination to stick files in.

cargo-screeps/src/copy.rs
daboross: If it should, it should change in both copy and upload. This preserves the current behavior of uploading files.

I.. think the purpose is to allow people to include extra javascript files if they want to?

Build will fail anyways if the project name is changed and `cargo clean` isn't run.

cargo-screeps/src/config.rs
ASalvail: Indeed, I had missed that. Nevermind!

daboross:
I think this should be fairly complete now!

There isn't backwards compatibility with the old configuration, but it shouldn't be hard to migrate - it's just adding an `[upload]` line. Instead `cargo screeps` will now warn on unused configuration values, and `cargo screeps upload` will complain if there isn't an `[upload]` section.

I'll update `screeps-starter-rust` after this is merged and released.

daboross:
Cool!

I've been trying it out locally too and it seems to be pretty reliable now. I'm not writing integration tests for this tool but I think that's probably alright, those can come later.

PR #35, opened 2018-09-09
Corrected sign_controller type.

ASalvail:
Fixes the type of `sign_controller` to target only `StructureController`.


daboross:
:+1:

PR #36, opened 2018-09-10
raw_memory improvements

ASalvail:
- Addresses some uniformity issues with the naming conventions.
- Sort some methods and add some documentation
- Add support for clearing the active foreign segment
- Fix typos in js code

### DIFF
--- a/screeps-game-api/src/raw_memory.rs
+++ b/screeps-game-api/src/raw_memory.rs
@@ -33,14 +33,6 @@ pub fn set_segment(id: i32, data: &str) {
     }
 }
 
-// pub fn get_foreign_segment() -> ForeignSegment {
-//     let v = js! {
-//         return RawMemory.foreignSegment;
-//     };
-    
-//     v.try_into().unwrap()
-// }
-
 get_from_js!(get_foreign_segment -> {
     RawMemory.foreignSegment
 } -> ForeignSegment);
--- a/screeps-game-api/src/raw_memory.rs
+++ b/screeps-game-api/src/raw_memory.rs
@@ -1,4 +1,12 @@
 //! Interface for Screeps `RawMemory` global object.
+#[derive(Deserialize, Debug)]
+pub struct ForeignSegment {
+    username: String,
+    id: String,
+    data: String
+}
+
+js_deserializable!(ForeignSegment);
 
 get_from_js!(get_active_segments -> { 
     Object.keys(RawMemory.segments).map(Number) 
@@ -25,17 +33,17 @@ pub fn set_segment(id: i32, data: &str) {
     }
 }
 
-get_from_js!(foreign_segment_username -> {
-    RawMemory.foreignSegment && RawMemory.foreignSegment.username
-} -> Option<String>);
+// pub fn get_foreign_segment() -> ForeignSegment {
+//     let v = js! {
+//         return RawMemory.foreignSegment;
+//     };
+    
+//     v.try_into().unwrap()
+// }
 
-get_from_js!(foreign_segment_id -> {
-    RawMemory.foreignSegment && RawMemory.foreignSegment.id
-} -> Option<i32>);
-
-get_from_js!(foreign_segment_data -> {
-    RawMemory.foreignSegment && RawMemory.foreignSegment.data
-} -> Option<String>);
+get_from_js!(get_foreign_segment -> {
+    RawMemory.foreignSegment
+} -> ForeignSegment);
 
 /// Implements `RawMemory.setActiveForeignSegment`
 /// 
--- a/screeps-game-api/src/raw_memory.rs
+++ b/screeps-game-api/src/raw_memory.rs
@@ -1,6 +1,6 @@
 //! Interface for Screeps `RawMemory` global object.
 
-get_from_js!(active_segments -> { 
+get_from_js!(get_active_segments -> { 
     Object.keys(RawMemory.segments).map(Number) 
 } -> Vec<i32>);
 
@@ -15,7 +15,7 @@ pub fn set_active_segments(ids: &[i32]) {
     }
 }
 
-get_from_js!(segment(id: u32) -> { 
+get_from_js!(get_segment(id: u32) -> { 
     RawMemory.segments[@{id}] 
 } -> Option<String>);
 
@@ -68,9 +68,9 @@ pub fn set_public_segments(ids: &[i32]) {
     }
 }
 
-get_from_js!(raw_memory -> {RawMemory.get()} -> String);
+get_from_js!(get -> {RawMemory.get()} -> String);
 
-pub fn set_raw_memory(value: &str) {
+pub fn set(value: &str) {
     js!{
         RawMemory.set(@{value});
     }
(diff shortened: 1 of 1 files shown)


screeps-game-api/src/raw_memory.rs
```diff
@@ -45,9 +68,9 @@ pub fn set_public_segments(ids: &[i32]) {
     }
 }
 
-get_from_js!(get() -> {RawMemory.get()} -> String);
+get_from_js!(raw_memory -> {RawMemory.get()} -> String);
```
daboross: Not necessarily against this rename, but I think it'd be a good idea to explicitly mention in a doc comment that this is `RawMemory.get` since the name won't match any more?

I'm not entirely sure `screeps::raw_memory::raw_memory()` is better than `screeps::raw_memory::get()` though, either? This name kind of [stutters](https://github.com/rust-lang-nursery/api-guidelines/issues/66).

I get it's slightly more consistent, but I think the consistency might be better enforced the other way, having it be `get` here and then having a `mod segments` module with similar `get`/`set` methods.

ASalvail: That indeed sounds like a better idea. Do you mind prefixing all "getters" with get?
Even the screeps API is pretty inconsistent with what is a "get" and what isn't.

daboross: I'd be good having `get`/`set` for all collection-like things that we can't just name equivalent to the screeps names. I don't think it would be worth it to prefix every other getter which doesn't have an equivalent setter with "get".

If we could, I'd prefer to keep all names exactly like the ones in the JS screeps api so they can be easily found and mentally mapped, but that doesn't really work for structures like `segments` which are just raw data able to be queried and updated.

I think `raw_memory` and `memory` are going to be the only two real places where it will make sense for us to have a basic 'set', and pairing that with 'get' seems semi-reasonable to me? 

----

Actually, hm, that brings up another possibility. We could implement a structure `RawMemorySegments` which wraps the `segments` raw and allows setting, getting and getting keys via methods. Or go more generic and make a `TypedJsObject<K, V>` wrapping `stdweb::Object` and have `segments()` return `TypedJsObject<i32, String>`.

What would you think of having something like that, and then having that allow us to keep screeps<->js names 100%? If we had typed object and list wrappers, we could even implement the methods on creep like `get_active_bodyparts` directly as `bodyparts()` returning a list reference.

ASalvail: > I think `raw_memory` and `memory` are going to be the only two real places where it will make sense for us to have a basic 'set', and pairing that with 'get' seems semi-reasonable to me?

I don't have a strong opinion either way. It just seemed weird to have the api with half of the methods called `get_*` and the other half just being `*`.

For the `TypedJSObject`, `stdweb` has conversion methods already for `HashMap<K, V>`, so we could make a type alias to a `HashMap<i32, ??>`. I suppose you might not want a string here given that rust's string are utf encoded, that might mess with what you expect it's length to be. Is there such a thing as a byte array that we could use? Maybe a `Vec<char>`?

For the typed wrapper, I suggested something similar and you told me it was too much overhead :P I was trying to come up with a midway alternative with objects wrapping references.

I started doing something in that direction, but we might have a chat on slack to throw ideas around and figure out what might work best. My main problem so far was _by far_ getting around the pesky `ConversionError`. Otherwise, most of the work is already done by `stdweb`.

ASalvail: Although, if you already have a clear idea in mind, please, be my guest.

The next thing I want to go through are the constants...

daboross: My main objection to using `HashMap` is that it loads all of the keys/values into rust memory. For something like `segments` this would be quite bad to do on every access- you could have a full 1MB of memory strings in that object, and access one should only require loading it's 100KB string into memory, not all of them. Same thing with `Vec`, though to a lesser extent.

`TypedJsObject` is what I mean by a typed wrapper here. I _think_ the typed wrapper you were mentioning was something more than that? Or if it wasn't, I was either mistaken or not thinking right.

ASalvail: Let's just say that my thoughts about the matter are evolving. I agree for the segments it'd be costly. However, I believe that for something like a creep, it might be worthwhile to require the whole, let's say, creep object, and destructure it (the fields, not the methods) in one go, rather than making multiple js calls.

Although, this might be trying to prematurely optimize things...

daboross: Yeah, I'm not at all sure about optimization either.

V8 should be able to optimize multiple JS calls pretty nicely as long as they're simple. My understanding of WASM<->JS interaction is that there isn't that much overhead _doing the function calls_, and the majority of cost of doing that is transferring data. For things like creep properties which are simple integers transferred on the stack, I can't imagine calling into JS isn't too much slower than just accessing the field would be in JS, and we'd need to do that field access for all the fields if we were pulling all of them in.

ASalvail: If you get the object that owns the field, isn't that as quick as getting one of their properties?

I would have imagined that the object is fetched in order to get one of their properties. Assuming (... yeah, that's a lot of assumptions) that those are contiguous in memory, fetching a whole object shouldn't be that much more costly than fetching a single id, _in JS_. Now, the transfer to WASM is another story.

Let's go with the simplest approach for now and optimize later if necessary.

daboross: Alright, simple is good.

I wouldn't be surprised if V8 did optimize structs and stick all the properties in one contiguous memory, but I also wouldn't assume it's going to do that, especially if the structure is only constructed once and used <5 times (like most screeps objects since they're reconstructed each tick).

screeps-game-api/src/raw_memory.rs
```diff
@@ -1,19 +1,16 @@
...
-    RawMemory.forignSegment && RawMemory.forignSegment.username
-} -> Option<String>);
-get_from_js!(foreign_segment_id -> {
-    RawMemory.forignSegment && RawMemory.forignSegment.id
-} -> Option<i32>);
-get_from_js!(foreign_segment_data -> {
-    RawMemory.forignSegment && RawMemory.forignSegment.data
-} -> Option<String>);
+
+js_deserializable!(ForeignSegment);
```
daboross: This looks nice!

I'm a bit sad about using json to transfer data back and forth but it does make sense to use it for simple stuff like this, and as you said we can tune and optimize it all later.

screeps-game-api/src/raw_memory.rs
```diff
@@ -26,13 +23,47 @@ pub fn set_active_segments(ids: &[i32]) {
...
+    }
+}
+
+// pub fn get_foreign_segment() -> ForeignSegment {
+//     let v = js! {
+//         return RawMemory.foreignSegment;
+//     };
+    
+//     v.try_into().unwrap()
+// }
```
daboross: If you're alright with it I think this'd be better just removed rather than commented out. It can stay in git history if we need it.

screeps-game-api/src/raw_memory.rs
ASalvail: Look at it as getting closer to a v1.0 and still having incentive to do better 😉

screeps-game-api/src/raw_memory.rs
ASalvail: Good catch, this wasn't meant to stay there.

PR #37, opened 2018-09-10
Implement HasId trait

ASalvail:


ASalvail:
Any ideas on how to reuse that for `get_object`?

Doing
```rust
pub fn get_object(id: &str) -> Option<impl HasId> {
    js_unwrap!(Game.getObjectById(@{id}))
}
```

seems pretty pointless.

ASalvail:
My best idea would be to move the declaration of `get_object` inside the `HasId` trait and basically use the method as a constructor of `Self`. My only problem with that approach is I don't know how to actually check whether it is actually the right object.

One possible solution to that problem, while not perfect, would be to hijack the ids and type them, maybe by appending something like `1134e143e...137243#Creep`.

screeps-game-api/src/objects/mod.rs
```diff
@@ -190,16 +190,22 @@ impl HasPosition for RoomPosition {
...
 
-/// All RoomObjects have positions.
-impl<T> HasPosition for T
-where
-    T: RoomObjectProperties,
-{
-    fn pos(&self) -> RoomPosition {
-        js_unwrap!(@{self.as_ref()}.pos)
+/// Trait covering all objects with an id.
+pub trait HasId : AsRef<Reference> {
```
daboross: What would you think of making this an `unsafe trait` to match the others?

It has a default method does relying on the behavior of `AsRef<Reference>::as_ref` returning something specific (something with an id method) like the other unsafe traits, and unlike `HasPosition`.

screeps-game-api/src/objects/mod.rs
```diff
@@ -217,6 +223,16 @@ pub unsafe trait RoomObjectProperties:
...
 
+/// All RoomObjects have positions.
+impl<T> HasPosition for T
+where
+    T: RoomObjectProperties,
+{
+    fn pos(&self) -> RoomPosition {
+        js_unwrap!(@{self.as_ref()}.pos)
+    }
+}
```
daboross: I *think* this would still belong directly under the definition for `HasPosition`?

daboross:
As far as I'm concerned I don't think we need to change `get_object` for this. `HasId` should be a strict subset of `RoomObjectProperties`, so I can't think of a much more useful thing to do than to return `RoomObject`.

I guess it means we can't immediately get the id from something which we've just gotten from an id, but that seems pretty niche.

Maybe we could add a `get_object_as_has_id` function which returns `Option<Box<HasId>>`?

screeps-game-api/src/objects/impls/construction_site.rs
```diff
@@ -3,17 +3,24 @@ use stdweb::unstable::TryInto;
...
 simple_accessors! {
     ConstructionSite;
-    (id -> id -> String),
     (my -> my -> bool),
     (progress -> progress -> i32),
     (progress_total -> progressTotal -> i32),
     (structure_type -> structureType -> StructureType),
 }
 
+impl PartialEq for ConstructionSite {
```
daboross: Do you think it would be at all reasonable to move automatic `PartialEq` / `Eq` implementations from `impl_structure` into an `impl_has_id` macro so that all things with ids could be covered?

I think things which implement `HasId` should be a strict superset of things implementing `StructureProperties`.

I guess it would mean we'd have to lay out every implementor which also implements from `StructureProperties` rather than having a blanket implementation, but it might be worth it to remove the need for duplicating these `PartialEq` implementations?

screeps-game-api/src/objects/mod.rs
daboross: I think it's safe to have this trait inherit from `RoomObjectProperties` too, since to my knowledge there is nothing with an ID which is not also a `RoomObject`.

daboross: I think this also deserves a `pub use` in `prelude` since it's a trait implementing functionality, and we want all regular methods on objects to be available as long as someone imports `screeps::prelude::*`.

ASalvail:
Actually, the more I'm thinking about it, the more I believe implementing type safe `TryFrom<Value>` for all rust objects, with js-side type check (ie using `instanceof`) would make the code much more readeable (and safe!) in the long run.

screeps-game-api/src/objects/mod.rs
ASalvail: The reasoning is that `RoomObjectProperties` isn't defined yet and its definition requires `HasPosition`.

How about we split `mod.rs` in more than one ginormous file (especially since there are still 2-3 other traits to add to the mix).

screeps-game-api/src/objects/impls/construction_site.rs
ASalvail: Agreed

screeps-game-api/src/objects/mod.rs
ASalvail: You are correct!

screeps-game-api/src/objects/mod.rs
daboross: We could definitely split it, but I think that'd be another PR. I think I'd rather keep trait implementations together with their source than have strict sequential definitions though, especially given that things can recursively depend on each other.

I've seen both putting things in order of least dependencies -> most dependencies as well as most dependencies -> least dependencies in different projects. I'm not really convinced either is as useful for code organization as keeping related things together is.

daboross:
Yeah, that could definitely be good. I think `stdweb` might even have some derive utilities to create `TryFrom` implementations for `Reference` if we want to do that.

The main reason I didn't implement them with type checks built in initially was the potential performance cost of doing an `instanceof` check every time rust grabs a reference. I don't know how much that performance cost is, but it seemed like it could be nontrivial, and it didn't seem worth it when we most of the time we can instead just trust that `Game.creeps` contains `Creep`s, for instance, or that `RoomObject.pos` is a `RoomPosition`.

I wouldn't necessarily be against changing `TryFrom` implementations to check for their source, but I'm still relatively confident I wouldn't want to perform those type checks when getting properties of things that we know the game always creates as the right type. Maybe we could have `TryFrom<Reference>` implementations do the checks and introduce a local `FromReferenceUnchecked` trait which a variant of `js_unwrap!` could use?

ASalvail:
That sounds like a best of both world approach. I'll take a look tonight.

ASalvail:
I just stumbled upon this: https://docs.rs/stdweb/0.4.6/stdweb/trait.ReferenceType.html

It'd fit the bill nicely for the quick, cheap conversion.

ASalvail:
This should address all your concerns.

ASalvail:
Regarding #51, I'm not yet convinced this is needed. I'd wait until we've used the API enough to make up our mind about it.

screeps-game-api/src/objects/mod.rs
```diff
@@ -119,6 +119,44 @@ where
     }
 }
 
+/// Trait covering all objects with an id.
+pub unsafe trait HasId: AsRef<Reference> + RoomObjectProperties {
```
daboross: I'd forgotten this before but I think `AsRef<Reference>` is now redundant because `RoomObjectProperties` also requires it.

screeps-game-api/src/objects/mod.rs
```diff
@@ -147,9 +185,6 @@ pub unsafe trait RoomObjectProperties:
 /// The reference returned by `AsRef<Reference>::as_ref` must reference a
 /// JavaScript object extending the `Structure` class.
 pub unsafe trait StructureProperties: RoomObjectProperties {
```
daboross: Since we don't have the blanket impl for all structures anymore, I think it would make sense to add `+ HasId` as a requirement for `StructureProperties` here.

Just to let someone require `T: StructuctureProperties` and know that `T` always has an ID as well.

daboross:
Thanks - everything's great!

Issue #30, opened 2018-09-09
Needs implementation


ASalvail:
A todo list of yet unimplemented API

- [x] `Game.market.*` (Transactions and Orders need testing)
- [x] `Game.market.incomingTransactions`
- [x] `Game.market.outgoingTransactions`
- [x] `Game.market.orders`
- [x] `Game.market.get_all_orders`
- [x] `Game.market.get_order`
- [x] `Game.notify`
- [x] `Game.map.findExit` (implement callback)
- [x] `Game.map.findRoute` (has callback)
- [x] `Game.getObjectById` (for objects that have an id, but do not implement `RoomObjectProperties`)
- [x] `Creep.body` (part of it is in `creep.parts`)
- [x] `Creep.saying`
- [x] `Creep.drop`
- [x] `Creep.moveByPath`
- [x] `Creep.moveTo` (implement `opts`)
- [x] `Creep.notifyWhenAttacked`
- [x] `Flag.*`
- [x] `Mineral.*`
- [x] `Nuke.*`
- [x] `Ressouce.id`
- [x] `Room.memory`
- [ ] `Room.visual` See #46
- [x] `Room.serializePath`
- [x] `Room.deserializePath`
- [x] `Room.findExitTo`
- [x] `Room.findPath`
- [x] `Room.getPositionAt`
- [x] `Room.lookAt`
- [x] `Room.lookAtArea`
- [ ] `RoomPosition.findClosestByPath` see #55 
- [x] `RoomPosition.findPathTo`
- [x] `RoomPosition.isEqualxy`
- [x] `RoomPosition.look`
- [ ] `RoomVisual.*` See #46 
- [x] `StructureController.activateSafeMode`
- [x] `StructureController.unclaim`
- [x] `StructureExtension.*` (I don't believe we can go with `HasStore`)
- [x] `StructureExtractor.*`
- [x] `StructureKeeperLair.*`
- [x] `StructureLab.*`
- [x] `StructureLink.*`
- [x] `StructureNuker.*`
- [x] `StructureObserver.*`
- [x] `StructurePowerBank.*`
- [x] `StructurePowerSpawn.*`
- [x] `StructurePortal.*`
- [x] `StructureRampart.*`
- [x] `StructureRoad.*`
- [x] `StructureSpawn.memory`
- [x] `StructureSpawn.spawnCreep` (options part, see todo in code)
- [x] `StructureSpawn.recycleCreep`
[...]

Issue #23, opened 2018-09-08
HasId trait + other Has* traits for common properties


daboross:
There are a number of things in Screeps which have ids and can be retrieved from `Game.getObjectById`. The set of things with ids is a subset of things implementing `RoomObject`.

It'd be nice to have a trait for this so that other code outside of `screeps-game-api` can be generic over things with IDs. I think the only reason we don't have one right now is because nothing inside the API bindings needed to be generic over that.

If this is considered a good idea, it might also be a good idea to split out `owner` / `my` into `Ownable`, and do the same for other common properties. `owner`/`my` are almost unique to `OwnedStructure`, but `Creep` and `ConstructionSite` also have them.

ASalvail:
We also have `HasStore` but we should add an `HasEnergy` for spawn, extensions, towers and rampart.

PR #38, opened 2018-09-10
Implement various methods.

ASalvail:
Best reviewed per commit.


ASalvail:
Addresses a lot of #30 points. 
What's left is everything that handles paths and a way to represent whatever `look` methods return.

Those will be separate PRs however, so this one should stand on its own.

screeps-game-api/src/constants.rs
```diff
@@ -830,3 +925,57 @@ impl TryFrom<Value> for ResourceType {
         })
     }
 }
+
+impl TryFrom<String> for ResourceType {
+    type Error = String;
+
+    fn try_from(s: String) -> Result<Self, Self::Error> {
```
daboross: See comment on `to_string`: I'd prefer to use integers to translate the various types across the rust<->js boundary whenever possible.

Not against having this implemented, but I think it'd be good to at least mention the str->int translations available on the JS side in a doc-comment here, and to warn about not using this if not necessary.

screeps-game-api/src/game.rs
```diff
@@ -170,12 +172,185 @@ pub mod map {
...
+    /// 
+    /// Does not yet support callbacks.
+    pub fn find_exit(from_room: Room, to_room: Room) -> Result<Exit, ReturnCode> {
+        let code: i32 = js_unwrap!{Game.map.findExit(@{from_room.name()}, @{to_room.name()})};
+        Exit::try_from(code).map_err(|v| v.try_into().expect("find_exit: Error code not recognized."))
+    }
+
+    #[allow(unused_variables)]
+    pub fn find_route(from_room: Room, to_room: Room, route_callback: Option<impl Fn(&str, &str) -> u32>) -> !{
+        unimplemented!()
```
daboross: I'd rather leave these methods out for the sake of having compile-time errors rather than runtime ones when trying to use them.

We could leave it in here commented out for the sake of keeping the signature, though.

Thoughts?

screeps-game-api/src/game.rs
```diff
@@ -170,12 +172,185 @@ pub mod map {
...
+    use {Room};
+    use constants::{ReturnCode, ResourceType};
+
+    pub enum OrderType {
+        Sell,
+        Buy
+    }
+
+    impl OrderType {
+        fn as_string(&self) -> String {
```
daboross: Since we are converting to a different type, `to_string` would be more appropriate than `as_string`.

See https://rust-lang-nursery.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv.

In addition, returning `&'static str` could be better here. Though if we do that, `as_str` might be a more appropriate name after all - it would indeed be free.

screeps-game-api/src/game.rs
```diff
@@ -170,12 +172,185 @@ pub mod map {
...
+        js_unwrap!{
+            Game.market.createOrder(@{order_type.as_string()},
+                                    __resource_type_num_to_str(@{resource_type as i32}),
+                                    @{price},
+                                    @{total_amount},
+                                    @{room.name()})
+        }
+    }
+
+    pub fn deal(order_id: &str, amount: u32, target_room: Room) -> ReturnCode {
```
daboross: Could this take `&Room` instead?

I'd almost want to have it taking a room name `&str` so that someone could execute deals without grabbing a reference to their room, but I don't feel too strongly about that.

screeps-game-api/src/objects/impls/creep.rs
```diff
@@ -38,10 +45,19 @@ impl Creep {
         js_unwrap!(@{self.as_ref()}.moveTo(@{x}, @{y}))
     }
 
+    #[allow(unused_variables)]
+    pub fn move_by_path(path: String) -> ! {
+        unimplemented!()
```
daboross: See above comment on `unimplemented!()`.

screeps-game-api/src/constants.rs
```diff
@@ -775,10 +821,59 @@ impl ResourceType {
             CatalyzedGhodiumAlkalide => 90,
         }
     }
+
+    pub fn to_string(&self) -> String {
```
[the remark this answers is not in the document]
ASalvail: Argh, I forgot about those!
I don't think having the string instead of the variant is useful. It was only meant to convert to a string before passing the hand to JS.

How about a wrapper around those JS functions that would implement `From<ResourceType> for Value::String` and `TryFrom<Value::String> for ResourceType` using those js methods instead of the usual conversion provided from stdweb?

screeps-game-api/src/game.rs
```diff
@@ -60,7 +60,7 @@ pub mod cpu {
     ///
     /// [http://docs.screeps.com/api/#Game.getHeapStatistics]: http://docs.screeps.com/api/#Game.getHeapStatistics
     ///
-    /// Returns object with all 0 values if heap statistics are not available.
+    /// Returns object with all 0 values if heap statistics are not availe.
```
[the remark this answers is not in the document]
ASalvail: Yup, thanks

screeps-game-api/src/game.rs
```diff
@@ -170,12 +172,185 @@ pub mod map {
 
     /// See [http://docs.screeps.com/api/#Game.map.isRoomAvailable]
     ///
-    /// [http://docs.screeps.com/api/#Game.map.isRoomAvailable]: http://docs.screeps.com/api/#Game.map.isRoomAvailable
+    /// [http://docs.screeps.com/api/#Game.map.isRoomAvaile]: http://docs.screeps.com/api/#Game.map.isRoomAvaile
```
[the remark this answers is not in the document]
ASalvail: 🤔 now those I think I corrected at some point, they might have reverted.... somehow 😕

screeps-game-api/src/game.rs
ASalvail: I like having them in the code because of the documentation generated (granted, this one doesn't have a doc).
I believed that using a function with the never return type wouldn't compile. After testing... it's annoying and confusing, so commenting it out seems like the best option.

screeps-game-api/src/game.rs
ASalvail: Would you rather we apply the same principle as for the resources and pass an integer instead? Again, this is solely used to call into JS, otherwise, the type is more than explicit enough.

screeps-game-api/src/game.rs
ASalvail: I do, it was a mistake on my part. Fixed!

screeps-game-api/src/objects/impls/structure_portal.rs
```diff
@@ -0,0 +1,34 @@
...
+    InterShard(InterShardPortalDestination),
+}
+
+
+impl StructurePortal {
+    pub fn destination(&self) -> PortalDestination {
+        let v = js! {return @{self.as_ref()}.destination;};
+
+        let is_inter_room: bool = js_unwrap!{
+            @{self.as_ref()}.destination instanceof RoomPosition
```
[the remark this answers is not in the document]
ASalvail: One would question why I even fetched the value if it weren't meant to be used 😆

screeps-game-api/src/objects/impls/structure_terminal.rs
```diff
@@ -0,0 +1,26 @@
...
+                                      @{amount},
+                                      @{destination})
+            },
+            Some(desc) => js_unwrap!{
+                @{self.as_ref()}.send(@{resource_type.to_string()},
+                                      @{amount},
+                                      @{destination},
+                                      @{desc})
+            },
+        }
```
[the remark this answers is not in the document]
ASalvail: It compiles, ~but I can't figure out why~. Nothing describes what an `Option` is translated as. Is `None` changed to `null` or to `undefined`? 

Turns out `None` becomes `Null`, but that has to go through the deserializer.

screeps-game-api/src/objects/mod.rs
```diff
@@ -225,12 +225,6 @@ pub unsafe trait RoomObjectProperties:
 /// The reference returned by `AsRef<Reference>::as_ref` must reference a
 /// JavaScript object extending the `Structure` class.
 pub unsafe trait StructureProperties: RoomObjectProperties {
-    fn hits(&self) -> i32 {
-        js_unwrap!(@{self.as_ref()}.hits)
-    }
-    fn hits_max(&self) -> i32 {
-        js_unwrap!(@{self.as_ref()}.hitsMax)
-    }
```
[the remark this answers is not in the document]
ASalvail: It's actually something I didn't think about since `Attackable` is implemented for all `StructureProperties` right now (I checked!), including `StructureController`.

Also, the situation is even worse than what you mentioned: calling `hits` on a controller returns `undefined`. This leads me to believe that the right solution might be to leave `hits` on `Attackable` and instead remove `Attackable` from `StructureController`. (To be fair, the game will let you attack it, but it's invincible.)

Your thoughts?

ASalvail:
Thanks a lot for taking the time to read through these.

I addressed most small things. I'll wait for your feedback on the points raised above before finishing the corrections.

screeps-game-api/src/game.rs
daboross: We could also have an `unimplemented-functions` feature which we enable when building docs but disable otherwise.

Or we could also change the function to accept a never/void input parameter '_unimplemented'! Never type in the return position can actually be called, it just guarantees the function will only ever panic or loop forever. If we include something like `enum UnimplementedParameterVoid {}` and then take `_unimplemented: UnimplementedParameterVoid` that would prevent anyone from calling it since no one would be able to construct that value.

screeps-game-api/src/constants.rs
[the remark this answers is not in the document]
daboross: The problem with `TryFrom<Value::String>` is that at that point the string has already been loaded into Rust, so it's equivalent to doing a method like this.

`Value::String` just wraps a rust `String`, not a pointer to a JS string. If we construct it or consume it at all we're already paying the cost of transfering the string between Rust and JS, which is what I want to avoid.

screeps-game-api/src/game.rs
daboross: If you think it's sane, sure!

I'm not 100% sure this optimization is the best idea, but then again it does seem worth it without much cost (just writing JS code to do the string->variant conversion instead of Rust code). I originally added it for resource types just because I imagined I would use those a lot. If it wouldn't be too much trouble an `__order_type_num_to_str` and `__order_type_str_to_num` pair seems like it'd be beneficial to me.

screeps-game-api/src/objects/impls/structure_terminal.rs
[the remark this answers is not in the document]
daboross: Turning into `null` or the value is what I'd expect `Option` to do, yeah. Glad it works!

screeps-game-api/src/objects/mod.rs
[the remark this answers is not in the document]
daboross: Oh- I had not realized `hits` on `StructureController` was undefined! It's there in the documentation and I assumed that was accurate. Thank you for the thorough investigation.

Being able to attack it in screeps makes sense, I guess. It's kind of loosely-constrained and I can see it not hurting them to include it. Wish they'd returned `0` from `hits` and `hitsMax`, but not everything can be perfect.

I think since we have `StructureProperties` and `Attackable` as separate types, we might as well make them as exact as we can. Let's go ahead and remove `Attackable` along with `hits` / `hitsMax` from `StructureController`?

There hasn't been any formal decision on whether we should try to make the Rust bindings more exact than the JS functionality, but it seems alright to me. Not sure if this is a slippery slope, but either way we can always change it later.

---

Since `StructureController` doesn't have `hits` and everything attackable does, keeping those properties on `Attackable` is alright with me.

Are you good with removing the `Attackable` implementation on `StructureController` and leaving the rest of this as it is?

daboross:
I'm glad to do it- thank you for doing this huge implementation pr!

I've replied to the individual threads.

screeps-game-api/src/game.rs
daboross: Sorry, I think that was a bit rambly/not clear.

I'd go ahead with this and apply the same principles. I'm not completely sure that it's the best idea, but I like it anyways, and I'm confident it will make the conversions faster.

screeps-game-api/src/objects/mod.rs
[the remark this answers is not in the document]
ASalvail: Until I get pull access on the repo, it's entirely your decision. 

Now if you want my opinion, I think we should correct the mistakes where possible as those are made glaring due to the type system. We already deviate from the standard API in many ways because we cannot have the type of function parameter representation they have in JS without sacrificing clarity. Same with using enums instead of constants.

Personnaly, I prefer clear and accurate over exact replica. Especially if you think a user of these rust bindings would have to correct type inaccuracies in their own code instead.

ASalvail:
There you go, that should address all of your concerns 😃

ASalvail:
Addresses most of #23

screeps-game-api/src/objects/mod.rs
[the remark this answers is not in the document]
daboross: Alright, glad to hear the input. Clear and accurate sounds good to me.

daboross:
LGTM! :+1: 

I think travis is complaining because of changes on the master branch so fixing that after merging seems sanest.

Opened https://github.com/daboross/screeps-in-rust-via-wasm/issues/52 for one lingering issue that I think could easily be dealt with separately.

screeps-game-api/src/game.rs
```diff
@@ -174,8 +176,182 @@ pub mod map {
...
+
+    #[repr(u32)]
+    #[derive(Clone, Debug)]
+    pub enum OrderType {
+        Sell = 0,
+        Buy = 1
+    }
+
+    // impl OrderType {
+    //     fn as_string(&self) -> String {
```
daboross: If you do want to keep this method in, I don't think it would hurt.

If we want to add in a typed `OrderType` to the transactions, the `serde` translation code would have to use strings anyways since `stdweb` turns it into JSON before loading it in.

PR #45, opened 2018-09-17
Switch `TryFrom` implementations to type check and add ExpectedType traits for non-typechecking conversions

daboross:
Should be pretty complete now.

Includes renaming `get_object` into `get_object_erased` and adding `get_object_typed` for attempting a cast before returning.

### DIFF
--- a/screeps-game-api/src/constants.rs
+++ b/screeps-game-api/src/constants.rs
@@ -4,10 +4,7 @@
 use stdweb::unstable::{TryFrom, TryInto};
 use stdweb::{Reference, Value};
 
-use {
-    objects::{FromExpectedType, RoomObject},
-    ConversionError,
-};
+use {objects::RoomObject, traits::FromExpectedType, ConversionError};
 
 enum_from_primitive! {
     #[repr(i32)]
@@ -323,7 +320,7 @@ pub mod look {
     use stdweb::unstable::TryInto;
 
     use {
-        objects::IntoExpectedType, ConstructionSite, Creep, Flag, Mineral, Nuke, Resource, Source,
+        traits::IntoExpectedType, ConstructionSite, Creep, Flag, Mineral, Nuke, Resource, Source,
         Structure, Terrain, Tombstone,
     };
 
--- a/screeps-game-api/src/constants.rs
+++ b/screeps-game-api/src/constants.rs
@@ -320,11 +320,10 @@ pub unsafe trait LookConstant {
 }
 
 pub mod look {
-    use stdweb::unstable::TryInto;
-
     use {
-        traits::IntoExpectedType, ConstructionSite, Creep, Flag, Mineral, Nuke, Resource, Source,
-        Structure, Terrain, Tombstone,
+        traits::{IntoExpectedType, TryInto},
+        ConstructionSite, Creep, Flag, Mineral, Nuke, Resource, Source, Structure, Terrain,
+        Tombstone,
     };
 
     use super::{Look, LookConstant};
--- a/screeps-game-api/src/constants.rs
+++ b/screeps-game-api/src/constants.rs
@@ -4,7 +4,10 @@
 use stdweb::unstable::{TryFrom, TryInto};
 use stdweb::{Reference, Value};
 
-use {objects::RoomObject, ConversionError};
+use {
+    objects::{FromExpectedType, RoomObject},
+    ConversionError,
+};
 
 enum_from_primitive! {
     #[repr(i32)]
@@ -63,7 +66,7 @@ impl TryFrom<i32> for ReturnCode {
 }
 
 pub unsafe trait FindConstant {
-    type Item: TryFrom<Value, Error = ConversionError> + AsRef<Reference>;
+    type Item: FromExpectedType<Reference>;
 
     fn find_code(&self) -> i32;
 }
@@ -309,7 +312,9 @@ pub enum Look {
 }
 
 pub unsafe trait LookConstant {
-    type Item: TryFrom<Value, Error = ConversionError>;
+    type Item;
+
+    fn convert_and_check_items(reference: Value) -> Vec<Self::Item>;
 
     fn look_code(&self) -> Look;
 }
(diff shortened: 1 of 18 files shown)


screeps-game-api/src/constants.rs
```diff
@@ -294,17 +299,18 @@ pub mod look {
...
+        RESOURCES, Look::Resources, Resource,  ::objects::IntoExpectedType::into_expected_type;
+        SOURCES, Look::Sources, Source,  ::objects::IntoExpectedType::into_expected_type;
+        MINERALS, Look::Minerals, Mineral,  ::objects::IntoExpectedType::into_expected_type;
+        STRUCTURES, Look::Structures, Structure,  ::objects::IntoExpectedType::into_expected_type;
+        FLAGS, Look::Flags, Flag,  ::objects::IntoExpectedType::into_expected_type;
+        CONSTRUCTION_SITES, Look::ConstructionSites, ConstructionSite,
+             ::objects::IntoExpectedType::into_expected_type;
+        NUKES, Look::Nukes, Nuke,  ::objects::IntoExpectedType::into_expected_type;
+        TERRAIN, Look::Terrain, Terrain,  ::stdweb::unstable::TryInto::try_into;
+        TOMBSTONES, Look::Tombstones, Tombstone,  ::objects::IntoExpectedType::into_expected_type;
```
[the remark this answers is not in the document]
daboross: True! The problem is that `::objects::IntoExpectedType::into_expected_type` and `::stdweb::unstable::TryInto::try_into` are pretty much the shortest for I can make them while still being able to differentiate between the two.

I suppose these could import `IntoExpectedType` and `TryInto` and leave out the full expanded form.

screeps-game-api/src/macros.rs
```diff
@@ -41,13 +41,40 @@
...
+/// let x: Creep = js!({ return Game.creeps.John; }).cast_expected_type().expect(...);
+/// ```
+///
+/// `cast_expected_type` will ensure that the return value is a [`stdweb::Reference`], but it won't
+/// do any more than that. If the JavaScript behaves incorrectly and returns something other than a
+/// Creep, and the `"check-all-casts"` feature is not enabled, it will silently make a
+/// [`screeps::Creep`] containing the wrong value which will fail when used.
+macro_rules! js_unwrap_ref {
+    ($($code:tt)*) => (
+        ::objects::IntoExpectedType::into_expected_type(js! { return $($code)* })
```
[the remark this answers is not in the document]
daboross: Sure- makes sense. I haven't been too strict since JS isn't but it does seem like a good idea.

screeps-game-api/src/game.rs
```diff
@@ -243,5 +243,5 @@ pub fn time() -> u32 {
 ///
 /// [http://docs.screeps.com/api/#Game.getObjectById]: http://docs.screeps.com/api/#Game.getObjectById
 pub fn get_object(id: &str) -> Option<::objects::RoomObject> {
-    js_unwrap!(Game.getObjectById(@{id}))
+    js_unwrap_ref!(Game.getObjectById(@{id}))
```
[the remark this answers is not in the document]
daboross: Once we make it generic- sure. This version still returns `RoomObject` though, and we can trust whatever is returned is indeed a `RoomObject`.

screeps-game-api/src/constants.rs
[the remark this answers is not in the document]
daboross: I've imported them in but I don't think there is more that I can do to make this nicer since we do need to differentiate between `IntoExpectedType` and `TryInto` for each one.

daboross:
@ASalvail Yeah, this definitely feels cleaner to me too.

I think this should address the points you've made- let me know if there's anything else you think needs attention. The comments might no longer be attached to the commits since I rebased the PR onto master but I think that should be alright.

screeps-game-api/src/constants.rs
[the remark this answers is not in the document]
ASalvail: You couldn't have made two arms to the macro? 
In the end, it doesn't matter, it gets expanded to the same code, it was just funny to see all the trouble you gave yourself 😛

screeps-game-api/src/game.rs
```diff
@@ -241,7 +245,24 @@ pub fn time() -> u32 {
 
 /// See [http://docs.screeps.com/api/#Game.getObjectById]
 ///
+/// This gets an object expecting a specific type and will return a `ConversionError` if the type
+/// does not match.
+///
+/// If all you want to assume is that something has an ID, use [`get_object_erased`].
+/// [http://docs.screeps.com/api/#Game.getObjectById]: http://docs.screeps.com/api/#Game.getObjectById
+pub fn get_object_typed<T>(id: &str) -> Result<Option<T>, ConversionError>
```
ASalvail: Not necessary for this PR (I can put it with `HasId`), but would you like having a wrapper type for IDs?

screeps-game-api/src/macros.rs
```diff
@@ -93,40 +120,48 @@ macro_rules! get_from_js {
...
+///     Room,
+///     // ...
 /// }
-/// 
+/// ```
+///
 /// Screeps game objects, in javascript, can be accessed via stdweb's `Reference`
 /// object. This macro: 
 ///   - Creates a struct named `objX`;
 ///   - Implements traits `AsRef<Reference>`, `TryFrom<Value>` for `objX`
```
ASalvail: Macro doc will require some adjustments.

ASalvail: Especially since the list of traits you get by deriving `ReferenceType` is far from obvious.
You get `AsRef<Reference>, `InstanceOf`, `TryFrom<Value>`, `TryFrom<Reference>, `[Try]From<T> for Reference/Value`...

Maybe just reference the doc for one of the object...

screeps-game-api/src/traits.rs
```diff
@@ -0,0 +1,72 @@
+//! Useful traits for interacting with JavaScript above what [`stdweb`] provides.
+use stdweb::unstable::TryFrom;
```
ASalvail: How about `pub use` both `TryFrom` and `TryInto` from _stdweb_ here? In the event of the trait stabilizing in rust, we could have a single line to change. Moreover, it'd make it possible for people to use both without needing to use `stdweb` directly, making the API more contained.

screeps-game-api/src/macros.rs
daboross: Right, will fix this. We might as well list everything `ReferenceType` does as well as mention it's all done through `#[derive(ReferenceType)]`.

screeps-game-api/src/traits.rs
daboross: Sounds like a good idea!

At first glance it seemed like it would be depending too much on an "unstable" API, but we're already using it everywhere and it's already a public dependency with our implementations on public types.

daboross:
This should address all comments @ASalvail.

ASalvail:
Looks great!

PR #48, opened 2018-09-18
Create CONTRIBUTING.md

daboross:
Seems like this would be a good idea. Nothing fancy.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,8 +14,11 @@ We adhere to the following guidelines:
 
 # Code Formatting
 
-The code formatting should match exactly with the behavior of `rustfmt` on nightly, though, so
-knowing all the intricacies of that shouldn't be necessary. To install `rustfmt`, use `rustup`:
+Please run `rustfmt` on the files you edit when submitting to this repository. This will handle all
+style issues mentioned in the 'fmt-rfcs' guidelines, so you should be able to simply run `cargo fmt`
+and be done with it.
+
+To install `rustfmt`, use `rustup`:
 
 ```console
 $ rustup component add --toolchain nightly rustfmt-preview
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,17 @@
+# Contributing to `cargo-screeps` and `screeps-game-api`:
+
+1. Ask questions. We're all in on the `#rust-wasm` channel on the [screeps slack].
```
ASalvail: > We're all in on the `#rust-wasm`

All-in? Or did you mean _We're all on the `#rust-wasm`_

daboross: Hm, yeah, the latter. It made sense at the time as in "we're all in [the place that is] on the `#rust-wasm` channel" but that's way less clear than saying "We're all in the `#rust-wasm` channel" or "We're all on the `#rust-wasm` channel".

daboross:
They're only a rough guide in most places, but they do have some good ideas!

The "Code Formatting" one should be all enforced by `rustfmt`, so it's slightly less interesting. I'll include that mention in there because I seem to have forgotten it.

CONTRIBUTING.md
```diff
@@ -12,6 +12,22 @@ We adhere to the following guidelines:
 - Code Formatting: [fmt-rfcs Rust Style Guide]
 - API building: [Rust API Guidelines]
 
+# Code Formatting
+
+The code formatting should match exactly with the behavior of `rustfmt` on nightly, though, so
```
ASalvail: Why not just outright require `rustfmt` to be ran on the code?
Also, I don't know what type of CI you have on the repo, but it might be integrated into it.

daboross: Requiring it be run sounds like a good idea.

The main reason I haven't added a CI check is just because of how fast `rustfmt` style can change. I haven't wanted to deny a PR just because some unrelated portion of code would be formatted differently in the newest version. If we're active enough, though, that could be avoided by regular "reformat" prs and requiring people to rebase.

ASalvail: I'm down for it. Let's ride the format train!

daboross: :+1: 

If we want to do the travis part I'll wait until we merge #40 to do that (or add that to #40).

PR #49, opened 2018-09-18
Use a new 'use' style

daboross:
This normalizes all `use` to a slightly different style that I've grown to like in other projects.

This also updates CONTRIBUTING.md (from #48) to include the logic of this use group style.

I mostly like this style because it groups uses together and creates no ambiguity as to how things should be formatted. There's always one `use` statement per top-level item, so there's never any decision that needs to be made about what imports to group together and what imports not to group together.

CC @ASalvail: see what you think of this.

I don't think it's a big thing, but it does affect all the code. I don't feel too strongly about this particular style, but I would like to normalize to something.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,9 @@ We adhere to the following guidelines:
 - Code Formatting: [fmt-rfcs Rust Style Guide]
 - API building: [Rust API Guidelines]
 
-# Code Formatting
+In addition, we have the following in-house guidelines:
+
+## Code Formatting
 
 Please run `rustfmt` on the files you edit when submitting to this repository. This will handle all
 style issues mentioned in the 'fmt-rfcs' guidelines, so you should be able to simply run `cargo fmt`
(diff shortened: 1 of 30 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+We adhere to the following guidelines:
+
+- Code Formatting: [fmt-rfcs Rust Style Guide]
+- API building: [Rust API Guidelines]
+
+In addition, we have the following in-house guidelines:
+
+## `use` formatting
+
+`use` statements should be at the top of the file, module or function they are used within.
```
ASalvail: Prior to module documentation?

CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+big enums, but it's best kept only within the function it's used in).
+
+When importing multiple things, `use` statements should be separated into the following
+newline-separated groups:
+
+- imports from `std`
+- imports from external crates
+- imports from the crate root
+- imports from `self` or `super`.
+
```
ASalvail: This follows the convention I'm used to, +1

ASalvail:
Might I also suggest alphebetical order when using collections of import?

CONTRIBUTING.md
daboross: Right, thanks for catching that. Meant after.

daboross:
Alphabetical order is good- rustfmt / the formatting guidelines should enforce that but it can't hurt to write it in here as well. I'll fix that and not mentioning module documentation


Re: Rust 2018 module system:

I've been avoiding using it for a while mostly because it's an unstable feature and there were some decisions yet to be made. I'd probably prefer to wait until it's stable to use, just to avoid extra work fixing any breaking changes that pop up. I realize we already depend on nightly, but still, the fewer different nightly things that break, the better. We can definitely switch when  the module system is stabilized and usable without another `#![feature]` attribute.

ASalvail:
That's fine by me. We can have a PR specifically to transition to rust 2018 later.

daboross:
I think this should address all of the comments and be up-to-date with master. We probably want to merge #40 first then this if both are good?

CONTRIBUTING.md
```diff
@@ -30,6 +32,66 @@ Then to format the code in this repository, use the following:
 $ cargo fmt
 ```
 
+## `use` formatting
+
+Private `use` statements should be above all other non-use items within the module or function they
```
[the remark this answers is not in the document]
daboross: :+1: I've added an order

daboross:
:+1:

thanks for reviewing and getting me to write all this down in the documentation!

PR #53, opened 2018-09-21
Implements Room.findPath()

ASalvail:
I'm very unsure of this one. For some obscure reasons (that might end up in our favor?!), `Room.findPath()` uses a mutable reference to the cost matrix to make its changes. I adapted the code in that direction, but kept to the `scoped_lts` approach. 

However, since a `CostMatrix` is received as input, it doesn't need lifetime annotations anymore. The thing is, I don't know enough about the intricacies of rust <-> wasm interractions to actually make a judgement call about whether this is still necessary.

I'm all ear to learn about it however 😄

### DIFF
--- a/screeps-game-api/src/constants.rs
+++ b/screeps-game-api/src/constants.rs
@@ -176,7 +176,7 @@ pub mod find {
 
 enum_from_primitive! {
     #[repr(i32)]
-    #[derive(Debug, PartialEq, Eq, Clone, Copy, Hash)]
+    #[derive(Debug, PartialEq, Eq, Clone, Copy, Hash, Deserialize)]
     pub enum Direction {
         Top = 1,
         TopRight = 2,
--- a/screeps-game-api/src/objects/impls/room.rs
+++ b/screeps-game-api/src/objects/impls/room.rs
@@ -18,9 +18,9 @@ use constants::{
     ReturnCode, 
     StructureType, 
 };
-use HasPosition;
 use memory::MemoryReference;
 use objects::{
+    HasPosition,
     Room, 
     RoomPosition,
     StructureController, 
--- a/screeps-game-api/src/objects/impls/room.rs
+++ b/screeps-game-api/src/objects/impls/room.rs
@@ -423,8 +423,8 @@ js_deserializable!{Step}
 
 #[derive(Debug, Deserialize)]
 pub enum Path {
-    VectorizedPath(Vec<Step>),
-    SerializedPath(String),
+    Vectorized(Vec<Step>),
+    Serialized(String),
 }
 
 js_deserializable!{Path}
--- a/screeps-game-api/src/objects/impls/room.rs
+++ b/screeps-game-api/src/objects/impls/room.rs
@@ -41,7 +41,7 @@ simple_accessors! {
     // todo: visual
 }
 
-scoped_thread_local!(static COST_CALLBACK: Box<(Fn(String, Reference) -> ())>);
+scoped_thread_local!(static COST_CALLBACK: Box<Fn(String, Reference) -> Option<Reference>>);
 
 impl Room {
     pub fn create_construction_site<T>(&self, at: T, ty: StructureType) -> ReturnCode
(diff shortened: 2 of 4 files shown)


ASalvail:
Correct me if I'm wrong, but the whole shenanigans with the closure that was needed for the `pathfinder::search` was due to the `CostMatrix` being generated from the callback, thus making it's lifetime hard to infer by the rust compiler.

In that case, implementing a callback for [`map.findRoute`](https://docs.screeps.com/api/#Game.map.findRoute) should not need any of that and simply use the callback functionalities provided by _stdweb_. Is that correct?

screeps-game-api/src/objects/impls/room.rs
```diff
@@ -99,6 +121,86 @@ impl Room {
...
+    pub fn find_path<O, T, F>(
+        &self, 
+        from_pos: &O, 
+        to_pos: &T,
+        opts: FindOptions<F>,
+    ) -> Vec<Step> 
+    where
+        O: HasPosition,
+        T: HasPosition,
+        F: Fn(String, CostMatrix) -> (),
```
daboross: The function should be able to return a `CostMatrix` in order to overwrite the one provided by the room- this probably wants to be `-> Option<CostMatrix>`.

screeps-game-api/src/constants.rs
```diff
@@ -176,7 +176,7 @@ pub mod find {
 
 enum_from_primitive! {
     #[repr(i32)]
-    #[derive(Debug, PartialEq, Eq, Clone, Copy, Hash)]
+    #[derive(Debug, PartialEq, Eq, Clone, Copy, Hash, Deserialize)]
```
daboross: Unless I'm mistaken this will deserialize via strings, not numbers. If it works, great.

If not, this will probably need a custom implementation which calls one of the methods to get from an i32.

screeps-game-api/src/objects/impls/room.rs
```diff
@@ -1,14 +1,34 @@
...
+use constants::{
+    Color, 
+    Direction,
+    find::Exit,
+    FindConstant, 
+    LookConstant, 
+    ReturnCode, 
+    StructureType, 
 };
+use HasPosition;
```
daboross: If you feel like cleaning up the use style, this should be in the `objects::` group. It's probably not correct anywhere else anyways though until #49 is merged.

screeps-game-api/src/objects/impls/room.rs
```diff
@@ -99,6 +121,86 @@ impl Room {
     //     unimplemented!()
     // }
 
+    pub fn find_path<O, T, F>(
+        &self, 
+        from_pos: &O, 
+        to_pos: &T,
+        opts: FindOptions<F>,
+    ) -> Vec<Step>
```
daboross: This only handles non-serialized paths.

Idea:

We could make `FindOptions` generic over "F: FindResultType", a trait like:

```rust
trait FindResultType {
    type Result;

    fn deserialize(v: Value) -> Result;
    fn serialized() -> bool;
}
```

then implement it for `RegularPath`, `SerializedPath` (or names you choose). This would let us return `F::Result` from `find_path` and handle both regular results which are `Vec<Step>`s and serialized results which are `String`s.

daboross:
Unfortunately the lifetime shenanigans are related to having a closure at all that we pass into JavaScript which is not `'static`. I think we'll need it any time we want to have full capability in callbacks.

ASalvail:
> Unfortunately the lifetime shenanigans are related to having a closure at all that we pass into JavaScript which is not `'static`. I think we'll need it any time we want to have full capability in callbacks.

In the closure example from [_stdweb_'s doc](https://docs.rs/stdweb/0.4.8/stdweb/index.html), none of this is mentioned. I don't believe that it the closure there is anymore static. I'm confused about whether it's a document failure on their end or if it's a problem particular to our use case that requires the whole unpacking of the callback.

screeps-game-api/src/objects/impls/room.rs
ASalvail: Honnestly, my first thought was not to support replacing the cost matrix. It is both inneficient and a pain to implement.

daboross:
Hm, you're right, it isn't mentioned there. The page doesn't explicitly mention that you _can_ use non-static data either, though. It just mentions "directly embedding rust code" which is fully doable with only static closures.

If you use non-static data in one of their examples, though, it will fail:

```rust
#[macro_use]
extern crate stdweb;

fn main() {
    let outer_owned_variable = "Goodbye.";
    let print_hello = |name: String| {
        println!("Hello, {}! {}", name, &outer_owned_variable);
    };

    js! {
        var print_hello = @{print_hello};
        print_hello( "Bob" );
        print_hello.drop(); // Necessary to clean up the closure on Rust's side.
    }
}
```
This errors with:

```console
$ cargo check
    Checking stdweb-test v0.1.0 (/home/daboross/stdweb-test)                                                                                                 
error[E0597]: `outer_owned_variable` does not live long enough                                                                                               
  --> src/main.rs:7:42                                                                                                                                       
   |                                                                                                                                                         
6  |     let print_hello = |name: String| {                                                                                                                  
   |                       -------------- value captured here                                                                                                
7  |         println!("Hello, {}! {}", name, &outer_owned_variable);                                                                                         
   |                                          ^^^^^^^^^^^^^^^^^^^^ borrowed value does not live long enough                                                  
...                                                                                                                                                          
15 | }                                                                                                                                                       
   | - `outer_owned_variable` dropped here while still borrowed                                                                                              
   |                                                                                                                                                         
   = note: borrowed value must be valid for the static lifetime...                                                                                           
                                                                                                                                                             
error: aborting due to previous error                                                                                                                        
                                                                                                                                                             
For more information about this error, try `rustc --explain E0597`.                                                                                          
error: Could not compile `stdweb-test`.                                                                                                                      

To learn more, run the command again with --verbose.
```

We could mirror the `stdweb` behavior and not accept closures with borrowed data. This would be fully doable, we could eliminate all of the extra code and greatly simplify PathFinder, but it would not allow closures to borrow local data. I made the decision to do all of the extra stuff because we know and trust the APIs we're using, and using borrowed data is generally a useful thing to do.

screeps-game-api/src/objects/impls/room.rs
ASalvail: Wouldn't it be easier just to make a `Path` enum with a `SerializedPath(String)` and a `VectorizedPath(Vec<Step>)` variant?

ASalvail:
Now that you mention it, I remember you telling me exactly this, but I suppose I'm a very forgetful person!

This is an amazing explanation however, we should put it in the documentation somewhere...

ASalvail:
Waiting for a second pass through the code.

screeps-game-api/src/constants.rs
daboross: The problem I'm thinking of is that direction constants actually are numbers in JS, not strings (like some of the other constants are).

See https://github.com/screeps/common/blob/f0214b373843fbe0e1d750108a5b8a1ebb34f1a9/lib/constants.js#L44 - they are defined as their integer numbers.

screeps-game-api/src/objects/impls/room.rs
daboross: It would be easier, but it then someone who only wants non-serialized paths would have to unwrap the result. Since it's in our power to give type-system-level assurance that the path is a given type, I'd rather do that than require extra runtime checks.

I mean it's fine if we don't handle non-serialized paths for now as long as we don't allow someone to set the 'serialized' FindOptions setting.

screeps-game-api/src/objects/impls/room.rs
```diff
@@ -168,3 +273,158 @@ impl PartialEq for Room {
...
+    x: u32,
+    y: u32,
+    dx: i32,
+    dy: i32,
+    direction: Direction
+}
+
+js_deserializable!{Step}
+
+#[derive(Debug, Deserialize)]
```
daboross: The `Deserialize` implementation for this will probably need a `#[serde(untagged)]` annotation since the names `VectorizedPath` and `SerializedPath` aren't part of the API return.

screeps-game-api/src/objects/impls/room.rs
```diff
@@ -168,3 +273,158 @@ impl PartialEq for Room {
...
+pub enum Path {
+    VectorizedPath(Vec<Step>),
```
daboross: `Path` in the variant names seems redundant here- it's already in the `Path` enum?

I _think_ `Path::Vectorized` would be just as clear as `Path::VectorizedPath`.

ASalvail: Very good idea.

screeps-game-api/src/objects/impls/room.rs
ASalvail: Good catch

daboross:
If you want, I could merge this PR as is and just merge #56 on top of it afterwards. Might be nice in order to have this with all the comments officially marked as "merged"?

ASalvail:
Please do, I'll finish #56 later

daboross:
:+1:

thanks!

PR #69, opened 2018-09-28
Implement Serialize and Deserialize for all constant structs.

daboross:
These implementations all use the real representations, not any of the invented numbers used in `TryFrom`. I'm hoping this won't cause confusion, but I'm not 100% sure.

My reasoning for adding these is for serializing/deserializing constants as part of larger structures separate from interacting with the API. I don't expect any of the `Serialize` or `Deserialize` implementations added to be used by `screeps-game-api` itself, but they could be used by consumers who want to serialize them with `serde-json` or `bincode` or some other `serde`-supported serialization format.

I have no idea why `Part` names are snake case whereas everything else is camel case...

### DIFF
--- a/screeps-game-api/Cargo.toml
+++ b/screeps-game-api/Cargo.toml
@@ -33,3 +33,4 @@ stdweb-derive = "0.5"
 
 [features]
 check-all-casts = []
+constants-serde = []
--- a/screeps-game-api/src/constants.rs
+++ b/screeps-game-api/src/constants.rs
@@ -11,7 +11,7 @@ use {
 
 enum_from_primitive! {
     #[repr(i32)]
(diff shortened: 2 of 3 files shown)


daboross:
That's a good point!

I'm not certain whether this can be effectively trimmed out or not. It'll definitely increase compile times and I can't be certain whether it will affect program size or not.

Maybe it'd be best to stick it under a `more-serde` feature?

daboross:
I've moved all the new implementations under a feature gate so they shouldn't have any affect if that isn't enabled.

daboross:
Alright- that sounds good to me. I do like features, but they can get unwieldy. We probably don't want to become like [`winapi`](https://github.com/retep998/winapi-rs/blob/0.3/Cargo.toml#L29-L373) with 300+ features controlling every aspect of the crate.

Issue #67, opened 2018-09-28
Tombstone has id method in addition to implementing `HasId`


daboross:
This should be removeable as `Tombstone` already implements `HasId`:

https://github.com/daboross/screeps-in-rust-via-wasm/blob/2623a57e440d40b2d0a69084c53dbd81d4254972/screeps-game-api/src/objects/impls/tombstone.rs#L8

PR #70, opened 2018-09-29
Bug patch: Tombstone Id and get_position_at

### LINKED ISSUE #66: getPositionAt incorrectly bound
I might be going a bit overboard on creating issues for things, but eh. If there are enough of them, that might justify a kanban board!

This code tries to call the JS function `get_position_at` whereas the function is called [getPositionAt](https://docs.screeps.com/api/#Room.getPositionAt):

https://github.com/daboross/screeps-in-rust-via-wasm/blob/0da7496ff2856b8974aa9b7ee93c179c200b4b58/screeps-game-api/src/objects/impls/room.rs#L105-L107

ASalvail:
Fixes #66 and #67 

Honestly, I think it'd have been quicker for you to patch it in github than to write the issue :P

### DIFF
--- a/screeps-game-api/src/objects/impls/room.rs
+++ b/screeps-game-api/src/objects/impls/room.rs
@@ -103,7 +103,7 @@ impl Room {
     }
 
     pub fn get_position_at(&self, x: u32, y: u32) -> Option<RoomPosition> {
-        js_unwrap!{@{self.as_ref()}.get_position_at(@{x}, @{y})}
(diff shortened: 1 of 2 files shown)


daboross:
:+1:

True! I was planning on just writing a patch soon too. The main reason I made issues was being in the middle of making other changes and not wanting to have to juggle the different git branches and getting distracted from the code I was working on finishing.

PR #71, opened 2018-09-29
September 2018 game update

ASalvail:
Implements the elements from the update described in [this changelog](https://blog.screeps.com/2018/09/changelog-2018-09-22/).

Except the sign string. This will be part of a review of all the constants that I plan to do a bit later.


screeps-game-api/src/objects/impls/room_terrain.rs
```diff
@@ -0,0 +1,15 @@
...
+    pub fn constructor(room_name: &str) -> Self {
+        js_unwrap!{new Room.Terrain(@{room_name})}
+    }
+
+    pub fn get(&self, x: u32, y: u32) -> Terrain {
+        js_unwrap!{@{self.as_ref()}.get(@{x}, @{y})}
+    }
+
+    pub fn get_raw_buffer(&self) -> Vec<u8> {
+        js_unwrap!(@{self.as_ref()}.getRawBuffer())
```
daboross: As mentioned in my comment, I think it should be pretty simple to create the `Vec<u8>` ourselves and ask the game to populate it rather than having the game create it and then copy it into WASM.

screeps-game-api/src/constants.rs
```diff
@@ -250,6 +250,9 @@ pub enum Terrain {
     Plain = 0,
     Wall = 1,
     Swamp = 2,
+    SwampWall = 3,
+    Lava = 4,
```
daboross: I don't *think* this should be necessary? I'm against including if if it doesn't occur just because it forces any consumers of `Terrain` to insert an extra case `Terrain::Lava => unreachable!(),` whenever matching on it which will never be triggered.

The screeps team has mentioned on slack that `TERRAIN_MASK_LAVA` has never been used, and that there are no current plans for using it. It's also missing from [Room.Terrain.get](https://docs.screeps.com/api/#Room.Terrain.get)'s listed output.

screeps-game-api/src/constants.rs
```diff
@@ -250,6 +250,9 @@ pub enum Terrain {
     Plain = 0,
     Wall = 1,
     Swamp = 2,
+    SwampWall = 3,
```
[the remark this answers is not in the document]
ASalvail: Oh, I didn't know that. Fair enough, I'll remove it.

screeps-game-api/src/objects/impls/room.rs
```diff
@@ -102,10 +104,19 @@ impl Room {
         }
     }
 
+    pub fn get_event_log(&self) -> Vec<Event> {
+        let raw_event_log: String = js_unwrap!{@{self.as_ref()}.getEventLog(true)};
```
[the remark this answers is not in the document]
ASalvail: I really don't see the usage, but sure.

screeps-game-api/src/objects/impls/room.rs
```diff
@@ -420,3 +431,120 @@ pub enum Path {
...
+    #[serde(flatten)]
+    pub event: EventType,
+    #[serde(rename = "objectId")]
+    pub object_id: String,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize)]
+#[serde(tag = "event", content = "data")]
+pub enum EventType {
+    #[serde(rename = "1")]
```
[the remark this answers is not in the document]
ASalvail: I have not tested it. My problem is working on getting those to actually make sense without, but I might have to write a deserializer manually for this one. I had forgotten that integers were a thing in JSON.

screeps-game-api/src/constants.rs
ASalvail: It's a question of whether you prefer to implement 1:1 the constants or clean-up the API on our end. I'm all for making the usage easier rather than going for 1:1 translation.

screeps-game-api/src/objects/impls/room.rs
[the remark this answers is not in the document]
ASalvail: It is actually being worked on right now: 
serde-rs/serde#1392

I might wait a few days and see whether this goes anywhere. Still, I'm learning how to manually write a deserializer in the meantime.

ASalvail:
So far, I'm unable to write a sane implementation of a deserializer for `Event`. Here's a summary of my problem, maybe you could have an idea.

The structure of the event log is defined as 
```js
{
    event: EVENT_ATTACK,
    objectId: '54bff72ab32a10f73a57d017',
    data: { /* ... */ }
}
```
with `data` depending on the `event` code, which is a `u8` ranging from 1 to 10. To represent this in rust, I have an `Event`
```rust
#[derive(Clone, Debug, PartialEq, Eq)]
pub struct Event {
    pub event: EventType,
    pub object_id: String,
}
```
and the inner `EventType`
```rust
#[derive(Clone, Debug, PartialEq, Eq, Deserialize)]
pub enum EventType {
    Attack(AttackEvent),
    ObjectDestroyed(ObjectDestroyedEvent),
    AttackController,
    Build(BuildEvent),
    Harvest(HarvestEvent),
    Heal(HealEvent),
    Repair(RepairEvent),
    ReserveController(ReserveControllerEvent),
    UpgradeController(UpgradeControllerEvent),
    Exit(ExitEvent),
}
```

Now, `serde_derive` provides an attribute `#[serde(tag = "event", container = "data")]`. This would work if the `event` field was _not_ a `u8`. Since this is not a possibility until the aforementioned PR is completed in serde, my option is to manually implement it.

```rust
impl<'de> Deserialize<'de> for Event {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        #[derive(Deserialize)]
        #[serde(field_identifier, rename_all = "camelCase")]
        enum Field { Event, ObjectId, Data };

        struct EventVisitor;

        impl<'de> Visitor<'de> for EventVisitor {
            type Value = Event;

            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
                formatter.write_str("struct Event")
            }

            fn visit_map<V>(self, mut map: V) -> Result<Event, V::Error>
            where
                V: MapAccess<'de>,
            {
                let mut event_type = None;
                let mut obj_id = None;
                let mut et = None;

                while let Some(key) = map.next_key()? {
                    match key {
                        Field::Event => {
                            if event_type.is_some() {
                                return Err(de::Error::duplicate_field("event"));
                            }
                            event_type = Some(map.next_value()?);
                        }
                        Field::ObjectId => {
                            if obj_id.is_some() {
                                return Err(de::Error::duplicate_field("objectId"));
                            }
                            obj_id = Some(map.next_value()?);
                        }
                        Field::Data => {
                            if et.is_some() {
                                return Err(de::Error::duplicate_field("data"));
                            }

                            let data: EventType;

                            if let Some(event_id) = event_type {
                                match event_id {
                                    // No clue
                                };
                            } else {
                                return Err(de::Error::custom("Found data before event."));
                            }

                            et = Some(data);
                        }
                    }
                }
                let et = et.ok_or_else(|| de::Error::missing_field("data"))?;
                let obj_id = obj_id.ok_or_else(|| de::Error::missing_field("objectId"))?;

                Ok(Event{event: et, object_id: obj_id})
            }
        }

        const FIELDS: &'static [&'static str] = &["event", "objectId", "data"];
        deserializer.deserialize_struct("Event", FIELDS, EventVisitor)
    }
}
```

This solution has two problems: 
- it assumes that the `event` field comes first. I couldn't figure out how to postpone the parsing of the value associated with the `data` key until the end of the parsing of the object. I could probably capture it in a `serde_json::Value`, but I have no clue how to destructure that later on.
- Even given the event code, I can't give a type hint to the parser to parse a specific variant. (e.g. cannot write `1 => data = map.next_value::<EventType::Attack>()?`.

So, short of interfacing the data with a `HashMap`, I am short on solutions. Any ideas?

screeps-game-api/src/constants.rs
daboross: In this case I would go for "cleanup", but I _think_ 1:1 would agree here too.  We're only ever returning the terrain, never passing it into the game, and the only place it's returned is documented to not return 4. Even though the constant exists, it's documented to never be used?

daboross:
I've been thinking about this and I'm not sure I have any better solutions. We can't always parse the tag first, since it won't always come first, and we need to do something with the data if it comes before.

The best solution I could think of is to deserialize the data itself into a `serde_json::Value` if it appears before the event type in the JSON. `serde_json::Value` does use a `HashMap` internally, but it has the advantage that it itself is a `serde::Deserializer` so we can deserialize data from it once we do have the event type.

If the event happens before the data, we can avoid that and deserialize the event directly.

I think this should be feasible? It would make the logic more complicated, but it should be able to work, and the custom `Deserialize` implementation you've written seems like it could be adapted to use this strategy.

As far as I know this is a limitation of how serde works. Deserializers read data in starting at the beginning and ending at the end and the thing being deserialized doesn't have much choice as to how it receives that input data. I would expect `serde-derive`'s implementation of externally-tagged data uses similar buffering, but I'm not sure.

Let me know what you think of this strategy.

screeps-game-api/src/objects/impls/room.rs
```diff
@@ -432,36 +436,148 @@ pub enum Path {
...
+                                            )))
+                                        }
+                                    };
+                                }
+                            };
+                        }
+                    }
+                }
+
+                if data_buffer.is_some() && event_type.is_some() && data.is_none() {
```
daboross: I think we should be able to remove the redundancy between these `is_some` calls and the `unwrap()` / `if let` below if we directly match on the values here. What would you think of something like this?

```rust
if data.is_none() {
    if let (Some(data_buffer), Some(event_type)) = (data_buffer, event_type) {
        ...
    }
}
```

screeps-game-api/src/objects/impls/room_terrain.rs
```diff
@@ -1,15 +1,37 @@
...
     }
 
     pub fn get_raw_buffer(&self) -> Vec<u8> {
-        js_unwrap!(@{self.as_ref()}.getRawBuffer())
+        let mut buffer: Vec<u8> = vec![0; 2500];
+        self.get_raw_buffer_array_ref(&mut buffer).expect("Panic in get_raw_buffer.");
+        buffer
     }
+
+    pub fn get_raw_buffer_array_ref<'a>(&self, buffer: &'a mut Vec<u8>) -> Result<&'a mut Vec<u8>, ReturnCode> {
```
daboross: Just bikeshedding here: what would you think of making this `get_raw_buffer_to_vec`?

Current name with the `_ref` suffix implies we'll be returning a cheap reference to something rather than making a new one, which is slightly misleading. There's a precedent for `to_xxx` when filling an existing buffer with bytes with functions like https://doc.rust-lang.org/std/io/trait.Read.html#method.read_to_string.

:+1: to having this as a public alternative to `get_raw_buffer` though!

screeps-game-api/src/objects/impls/room_terrain.rs
```diff
@@ -1,15 +1,37 @@
...
+        let is_success: bool;
+        {
+            let arr: UnsafeTypedArray<u8> = unsafe{ UnsafeTypedArray::new(buffer.as_mut_slice()) };
+
+            is_success = js!{
+                var bytes = @{arr};
+                return @{self.as_ref()}.getRawBuffer(bytes) != -10;
```
daboross: Could we possibly unwrap a `ReturnCode`, or at least compare to `0` rather than `-10`? Documentation states we'll only get `-10` but that could change in the future and the server doesn't always follow the spec. I think it would be better to err on the side of reporting failure rather than seeing any non-`-10` error code as a success.

daboross: Unrelated to other concerns:

Do you think it would be reasonable to `assert_eq!(bytes.len(), 2500);` before we call into JS or should we assume that `getRawBuffer` will error if we give it an incorrect length?

screeps-game-api/src/objects/impls/room.rs
```diff
@@ -420,3 +435,228 @@ pub enum Path {
...
+                    let val: serde_json::Value = data_buffer.unwrap();
+                    let err = |_| de::Error::custom("Can't parse Event data.");
```
daboross: Could we include the display representation of the original error here? It'd be much more useful than a static string. Just something like `de::Error::custom(format_args!("can't parse event data due to inner error {}", e))` could work? We should also avoid capitalization and an ending period as serde inserts this as part of a larger sentence (or at least that's what [custom](https://docs.rs/serde/1.0.79/serde/de/trait.Error.html#tymethod.custom)'s documentation states?).

screeps-game-api/src/objects/impls/room_terrain.rs
```diff
@@ -0,0 +1,37 @@
...
+            }.try_into().unwrap();
+        }
+        if is_success {
+            Ok(buffer)
+        } else {
+            Err(ReturnCode::InvalidArgs)
+        }
+    }
+
```
daboross: I think this blank line is a formatting error? I'm 70% sure `rustfmt` would remove it but not more than that.

screeps-game-api/src/objects/impls/room.rs
ASalvail: I had avoided doing that because I didn't want to get 3 levels deep, but that fixes my concern nicely. Looks great!

daboross: Cool!

We might even be able to do 
```rust
if let (Some(data_buffer), Some(event_type), None) = (data_buffer, event_type, data) {
```

but I'm not 100% sure rust will allow that since we're modifying `data` inside and lifetimes are lexical (can't wait for non-lexical lifetimes but that's a whole other topic).

screeps-game-api/src/objects/impls/room_terrain.rs
ASalvail: I can't compare with zero as success is defined as returning the array. I could potentially compare it with `bytes` to make sure it's the same ref? I'll have to look this up though.

ASalvail: I was going with the second. The error code should be exactly for that purpose, so checking here would be duplicating the effort.

ASalvail: On the other hand, we could check in our function and panic on any error as they're not supposed to happen.

screeps-game-api/src/objects/impls/room_terrain.rs
ASalvail: I swear I run it, but I think VSCode doesn't like me. I'll use cargo instead in the future...

screeps-game-api/src/objects/impls/room_terrain.rs
ASalvail: A quick look seems to indicate that it'd compare all 2500 values :/

ASalvail:
No need to be sorry, I can give as much as I take :P

PR #76, opened 2018-10-08
findRoute and findExit callback implementations

ASalvail:
Use the same `scoped_thread_local` container to work both callbacks.

Both functions have a regular and a `*_callback" version.

Also adds a `game::map::ExitDirection` struct to express the output of `find_route`.

### DIFF
--- a/screeps-game-api/src/game.rs
+++ b/screeps-game-api/src/game.rs
@@ -127,7 +127,9 @@ pub mod gcl {
 ///
 /// [http://docs.screeps.com/api/#Game.map]: http://docs.screeps.com/api/#Game.map
 pub mod map {
-    use std::collections;
+    use std::{collections, mem};
+
+    use stdweb::Value;
 
     use {
         constants::{find::Exit, Direction, ReturnCode},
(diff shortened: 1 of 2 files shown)


screeps-game-api/src/game.rs
```diff
@@ -186,17 +188,80 @@ pub mod map {
     }
 
     /// Implements `Game.map.findExit`.
-    ///
-    /// Does not yet support callbacks.
-    pub fn find_exit(from_room: Room, to_room: Room) -> Result<Exit, ReturnCode> {
-        let code: i32 = js_unwrap!{Game.map.findExit(@{from_room.name()}, @{to_room.name()})};
+    pub fn find_exit(from_room: &Room, to_room: &Room) -> Result<Exit, ReturnCode> {
```
daboross: Just realized this - sorry for not thinking of this when `find_exit` was originally added: we should allow accepting room names here as well as or instead of `Room` references so we can find exits to rooms not visible.

This is more important for `findRoute` which is much more often used to find routes to far away rooms than it is for `findExit`. Both should be able to accept room names, though, since `Room` objects won't exist for rooms not visible.

My best idea for solving this would be to just accept `&str`, but if you want we could also add a `HasRoomName` trait or something similar implemented both for `Room` and `String` / `str`.

screeps-game-api/src/game.rs
```diff
@@ -236,29 +236,34 @@ pub mod map {
...
 
-    fn parse_find_route_returned_value(v: Value) -> Result<Vec<ExitDirection>, ReturnCode> {
+    fn parse_find_route_returned_value(v: Value) -> Result<Vec<RoomRouteStep>, ReturnCode> {
         match v {
             Value::Number(x) => {
                 let i: i32 = x.try_into().unwrap();
-                Err(i.try_into().expect("Unexpected return code."))
+                Err(i
+                    .try_into()
+                    .expect(&format!("Unexpected return code: {:?}.", i)))
```
daboross: I think if we want to include a message here, it would be better to `match` on the result or use `.unwrap_or_else(|_| panic!())` rather than use `.expect()`. Using `format!()` like this will allocate a new string and do string formatting even if the result is `Ok()`.

ASalvail:
That should be good to go :)

daboross:
Thanks! :+1:

PR #84, opened 2018-10-10
Configurable initialization header for cargo-screeps

daboross:
Depends on #85.

Previously, `cargo-screeps` generated a fixed header in `main.js` for initializing the WASM module.

This allows changing that header on a per-project basis so as to allow including things at the user's discretion. I'm imagining this could be commonly used for things like not trying to load the WASM module if a user's bucket is below 500, or measuring exact CPU usage of each step loading the WASM.

Another use case would be integrating a rust wasm module into another project. We already allow changing the file name from `main.js` to something else, this would also allow custom initialization, for example to export things out to other modules.

## Functional changes

In addition to allowing customization, this changes the default initialization to cache the WASM module, reusing it if later initialization fails. I discovered the need for this when recently trying out a project and finding initialization timing out. The WASM module cached has no state, it's purely compiled code, so it should be completely safe.

### DIFF
--- a/cargo-screeps/README.md
+++ b/cargo-screeps/README.md
@@ -110,6 +110,57 @@ This configures general build options.
 - `output_js_file`: the javascript file to export bindings and bootstrapping as
   (default `"main.js"`)
 - `output_wasm_file`: the WASM file to rename compile WASM to (default `"compiled.wasm"`)
+- `initialize_header_file`: a file containing the JavaScript for starting the WASM instance. See
+  [overriding the default initialization header](#overriding-the-default-initialization-header)
+
+## Overriding the default initialization header
+
+`cargo-screeps` tries to make a reasonable `main.js` file to load the WASM. However, it's pretty
+basic, and you might find you want to do some things in JavaScript before loading the WASM module.
+
+Luckily, you can override this initialization! Set `build.initialize_header_file` to a file containing the JavaScript initialization code.
+
+The code will have access to two utility functions, `wasm_fetch_module_bytes` and `wasm_create_stdweb_vars`:
+
+```js
+/**
+ * Fetches WASM bytes for the wasm module.
+ *
+ * These should be given to `new WebAssembly.Module` to instantiate a WebAssembly module.
+ */
+function wasm_fetch_module_bytes() { /* ... */ }
+
+/**
+ * Creates the stdweb wrapper for a module instance.
+ *
+ * It has two properties, `imports` and `initialize`. `imports` should be passed as the second
+ * argument to `new WebAssembly.Instance` to provide the WASM module with imports, and `initialize`
+ * should be called on the resulting WebAssembly instance.
+ *
+ * Calling `initialize` will finish associating the imports with the wasm module, and will call the
+ * rust module's main function.
+ */
+function wasm_create_stdweb_vars() { /* ... */ }
+```
+
+When you override the initialization, you will need to perform it yourself. Here is the default initialzation header for reference:
+
+```js
+"use strict";
+let wasm_module = null;
+
+function wasm_initialize() {
+    if (wasm_module == null) {
+        let wasm_bytes = wasm_fetch_module_bytes();
+        wasm_module = new WebAssembly.Module(wasm_bytes);
+    }
+    let stdweb_vars = wasm_create_stdweb_vars();
+    let wasm_instance = new WebAssembly.Instance(wasm_module, stdweb_vars.imports);
+    stdweb_vars.initialize(wasm_instance);
+}
+
+module.exports.loop = wasm_initialize;
+```
 
 # Updating `cargo screeps`
 
--- a/cargo-screeps/README.md
+++ b/cargo-screeps/README.md
@@ -118,51 +118,13 @@ This configures general build options.
 `cargo-screeps` tries to make a reasonable `main.js` file to load the WASM. However, it's pretty
 basic, and you might find you want to do some things in JavaScript before loading the WASM module.
 
-Luckily, you can override this initialization! Set `build.initialize_header_file` to a file containing the JavaScript initialization code.
+Luckily, you can override this initialization! Set `build.initialize_header_file` to a file
+containing the JavaScript initialization code.
 
-The code will have access to two utility functions, `wasm_fetch_module_bytes` and `wasm_create_stdweb_vars`:
+Two utility functions `wasm_fetch_module_bytes` and `wasm_create_stdweb_vars` will always be
+created, but the initialization header controls what actually runs.
 
-```js
-/**
- * Fetches WASM bytes for the wasm module.
- *
- * These should be given to `new WebAssembly.Module` to instantiate a WebAssembly module.
- */
-function wasm_fetch_module_bytes() { /* ... */ }
-
-/**
- * Creates the stdweb wrapper for a module instance.
- *
- * It has two properties, `imports` and `initialize`. `imports` should be passed as the second
- * argument to `new WebAssembly.Instance` to provide the WASM module with imports, and `initialize`
- * should be called on the resulting WebAssembly instance.
- *
- * Calling `initialize` will finish associating the imports with the wasm module, and will call the
- * rust module's main function.
- */
-function wasm_create_stdweb_vars() { /* ... */ }
-```
-
-When you override the initialization, you will need to perform it yourself. Here is the default initialzation header for reference:
-
-```js
-"use strict";
-let wasm_module = null;
-
-function wasm_initialize() {
-    if (wasm_module == null) {
-        let wasm_bytes = wasm_fetch_module_bytes();
-        wasm_module = new WebAssembly.Module(wasm_bytes);
-    }
-    let stdweb_vars = wasm_create_stdweb_vars();
-    let wasm_instance = new WebAssembly.Instance(wasm_module, stdweb_vars.imports);
-    stdweb_vars.initialize(wasm_instance);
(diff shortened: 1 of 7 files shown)


cargo-screeps/README.md
```diff
@@ -46,7 +46,114 @@ This is not fully tested, but feel free to use! Issues are welcome.
...
+  [overriding the default initialization header](#overriding-the-default-initialization-header)
+
+## Overriding the default initialization header
+
+`cargo-screeps` tries to make a reasonable `main.js` file to load the WASM. However, it's pretty
+basic, and you might find you want to do some things in JavaScript before loading the WASM module.
+
+Luckily, you can override this initialization! Set `build.initialize_header_file` to a file containing the JavaScript initialization code.
+
+The code will have access to two utility functions, `wasm_fetch_module_bytes` and `wasm_create_stdweb_vars`:
```
ASalvail: Wouldn't that part be much easier to understand if you made a template or maybe just commented `default_initialization_header`?

cargo-screeps/README.md
```diff
@@ -46,7 +46,114 @@ This is not fully tested, but feel free to use! Issues are welcome.
...
+ *
+ * Calling `initialize` will finish associating the imports with the wasm module, and will call the
+ * rust module's main function.
+ */
+function wasm_create_stdweb_vars() { /* ... */ }
+```
+
+When you override the initialization, you will need to perform it yourself. Here is the default initialzation header for reference:
+
+```js
```
ASalvail: For instance, you could simply point out what is essential here and what isn't. Where could I insert code and when would it be called?

It might also be just my js ignorance talking...

cargo-screeps/src/build.rs
```diff
@@ -188,65 +175,105 @@ if( typeof Rust === "undefined" ) {
...
-(function( factory ) {{
-    // stripped for screeps usage
-    factory();
-}}( function() {{
-    return (function( module_factory ) {{
-        // replaced with hardcoded grab for screeps usage
-        var instance = module_factory();
+        r#"{}
+
+/**
```
ASalvail: Could the following also be extracted in their own template rather than hardcoded here?

Also, I'm not sure what I think of the comments in the js file, especially hardcoded ones. I know we are talking about a few bytes, but this is still bytes that could be used for code rather than for comments.

daboross: Yeah, comments could definitely be removed. If we're that pressed for bytes, though, running a minimizer on main.js would remove these comments and more.

The reason this section is hardcoded right now is that it's so closely tied to `cargo-web` version. Particularly if the output format changes, we'll need to change both the prefix/suffix patterns and these functions to correctly get the right part of stdweb's wrapper.

We could pull it out into a template, but I don't see any need for customization, and we would have two or more places to modify when updating to new `cargo-web` changes. Then, if a user has customized it, they're customizations will break when updating `cargo-web` and `cargo-screeps`. I'm hoping by keeping this as part of `cargo-screeps` we can make it the more stable interface that further customizations can build on.

cargo-screeps/README.md
daboross: :+1: this is a good idea. I'll write a more in-depth guide to address this and to give a better idea of customizations.

cargo-screeps/README.md
daboross: I'm not sure? I mean I wrote this section with the intention of someone who knows the WebAssembly JavaScript types. I thought that giving a clear idea of what someone writing a custom function has access to would be better than _only_ example code.

I can definitely make more documentation here and write more example code, better commented, but I think showing the functions one can call provides value that well-commented example code doesn't.

daboross:
Thanks for looking at this!

I'll go over and write more/better documentation for this soon.

---

There's one other thing I'm wary about here. Right now, in this PR, `initialization_header_file` is stuck in with `screeps.toml`, and all other configuration options in that file are somewhat sensitive / per-user. I imagine `initialization_header_file`, on the other hand, would want to be the same for a project regardless of who's deploying it: JavaScript code inlined in rust projects can rely on the functions the initialization header provides, and rely on its behavior.

I'm considering splitting `screeps.toml` into `screeps-deploy.toml` and `screeps-build.toml` for this reason. `screeps-deploy.toml` would be what `screeps.toml` currently is - it would be kept out of version control and configured per deploying user. `screeps-build.toml` would be checked in to version control and managed like `Cargo.toml` and other source files.

cargo-screeps/README.md
ASalvail: Of course, I think I meant "in addition of".

ASalvail:
I can see the value in your suggestion, even though I think this would have been only a slight annoyance.

However, your comment does make me realize that you didn't change `screep-default.toml` with the new parameter.

daboross:
> I can see the value in your suggestion, even though I think this would have been only a slight annoyance.

I think I'm starting to agree more with this - that it wouldn't be too valuable. Projects can already have `screeps-default.toml`, and that might as well contain these settings.

I've made some doc improvements and rebased, I'll push those updates now.

cargo-screeps/docs/initialization-header.md
```diff
@@ -0,0 +1,136 @@
...
+        // Fetch the source byte array. `wasm_fetch_module_bytes` just returns
+        // require('wasm_module_name'), but it's useful to have as a separate function
+        // so we don't hardcode the module name.
+        let wasm_bytes = wasm_fetch_module_bytes();
+        // Compile the module. I've found that this can time out, but because of however it works,
+        // `wasm_module` will still be initialized.
+        wasm_module = new WebAssembly.Module(wasm_bytes);
+    }
+    // Initialize `stdweb`. This returns an object with an `initialize` function and `imports`
+    // object, meant to be used as bellow:
```
ASalvail: bellow -> below

cargo-screeps/docs/initialization-header.md
```diff
@@ -0,0 +1,136 @@
...
+    module.exports.loop();
+}
+
+module.exports.loop = wasm_initialize;
+```
+
+For reference, see [WebAssembly.Module] and [WebAssembly.Instance] docs.
+
+The two utility functions, `wasm_fetch_module_bytes` and `wasm_create_stdweb_vars` will always be
+generated, regardless of initialization header. Their purpose is to provide an interface to stdweb's internal initialization, and to remain stable between stdweb and cargo-web updates.
```
ASalvail: This seems like an unusually long line.

cargo-screeps/docs/initialization-header.md
```diff
@@ -0,0 +1,136 @@
...
+    // object, meant to be used as bellow:
+    let stdweb_vars = wasm_create_stdweb_vars();
+    // Create the WebAssembly instance.
+    let wasm_instance = new WebAssembly.Instance(wasm_module, stdweb_vars.imports);
+    // Run the remaining stdweb initialization. This is fairly short, just some things to make sure
+    // callbacks work right and such. This will also run the rust `main` function.
+    stdweb_vars.initialize(wasm_instance);
+    // Now we run the actual main loop - this assumes that the rust `main` function overwrites
+    // `module.exports.loop`.
+    module.exports.loop();
```
ASalvail: Isn't that equivalent to recursively calling `wasm_initialize`? Or does that bind beforehand and now it basically injects `wasm_initialize` before `module.exports.loop()` is called?

cargo-screeps/docs/initialization-header.md
```diff
@@ -0,0 +1,136 @@
...
+    }
+    // Initialize `stdweb`. This returns an object with an `initialize` function and `imports`
+    // object, meant to be used as bellow:
+    let stdweb_vars = wasm_create_stdweb_vars();
+    // Create the WebAssembly instance.
+    let wasm_instance = new WebAssembly.Instance(wasm_module, stdweb_vars.imports);
+    // Run the remaining stdweb initialization. This is fairly short, just some things to make sure
+    // callbacks work right and such. This will also run the rust `main` function.
+    stdweb_vars.initialize(wasm_instance);
+    // Now we run the actual main loop - this assumes that the rust `main` function overwrites
```
ASalvail: main -> game. I got confused about why the `main` from rust was called multiple time.

cargo-screeps/docs/initialization-header.md
daboross: The second, if I understand correctly.

`stdweb_vars.initialize` will call the `fn main()` in rust code, and it's assumed that that function does `module.exports.loop = @{some_rust_fn};`.

So- yes, this is essentially injecting `wasm_initialize` before the regular `loop` function.

cargo-screeps/docs/initialization-header.md
daboross: Could you explain more?

I really do mean the `main` function here. Like, `fn main() {` in `src/main.rs`, written in rust.

ASalvail: But as far as I understand it, the `module.exports.loop` function at this point points to the rust `game_loop` (in terms of the rust starter script) not the `main` function from the rust code. Isn't that right?

ASalvail: As in 
> Now we run the actual ~main~ game loop

daboross: Oh! Right. Definitely- I misunderstood.

daboross:
I think my documentation is a bit imperfect, but probably better than nothing given the functionality increase this offers. Going to merge this and hope we can improve it and make it more clear in future PRs?

PR #86, opened 2018-10-14
Implements SpawnOptions and Spawning

ASalvail:
I modified a little bit how `SpawningOptions` worked to be more similar to the other builders we have peppered throughout the library. At the same time, I made it so it is actually supported via a `spawn_creep_with_options` method.

I also added a `Spawning(Reference)` that covers the `StructureSpawn.spawning` accessor. I was mostly done when I realized it may have made more sense to just put all the accessors within the `StructureSpawn` reference, but as we try to emulate the base API as closely as possible, I figured this would fit best.

### DIFF
--- a/screeps-game-api/src/objects/mod.rs
+++ b/screeps-game-api/src/objects/mod.rs
@@ -87,6 +87,8 @@ reference_wrappers!(
     StructureRoad,
     #[reference(instance_of = "StructureSpawn")]
     StructureSpawn,
+    #[reference(instance_of = "Spawning")]
+    Spawning,
     #[reference(instance_of = "StructureStorage")]
     StructureStorage,
     #[reference(instance_of = "StructureTerminal")]
--- a/screeps-game-api/src/objects/mod.rs
+++ b/screeps-game-api/src/objects/mod.rs
@@ -295,8 +295,7 @@ pub unsafe trait CanStoreEnergy: StructureProperties {
 ///
 /// # Contract
 ///
-/// The reference returned from `AsRef<Reference>::as_ref` must be have an
-/// `energy` and an `energyCapacity` properties. Also, that energy can be used
+/// The reference returned from `AsRef<Reference>::as_ref` must be able to be used
 /// by a spawner to create a new creep.
 pub unsafe trait HasEnergyForSpawn: CanStoreEnergy {}
 
--- a/screeps-game-api/src/objects/mod.rs
+++ b/screeps-game-api/src/objects/mod.rs
@@ -289,6 +289,15 @@ pub unsafe trait CanStoreEnergy: StructureProperties {
     }
 }
 
+/// Used to specify which structures can use their stored energy for spawning creeps.
+///
+/// # Contract
+///
+/// The reference returned from `AsRef<Reference>::as_ref` must be have an
+/// `energy` and an `energyCapacity` properties. Also, that energy can be used
+/// by a spawner to create a new creep.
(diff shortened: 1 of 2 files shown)


daboross:
Looks good!

I'm a bit sad to have

```rust
structure_spawn.spawn_creep_with_options(
    &[MOVE, CARRY, WORK]
    name,
    SpawnOptions::new()
        .memory(&memory)
        .directions(&directions)
)
```

rather than

```rust
structure_spawn
    .spawn(&[MOVE, CARRY, WORK], &name)
    .memory(&memory)
    .directions(&directions)
    .execute()
```

but I suppose the former is more consistent with other builder patterns...

Looking at `SpawnOptions`, it seems I didn't actually build it in the builder style which would allow chaining like `.memory(&memory).directions(&directions).execute()`... That was certainly the intention. I have no idea how I missed that and didn't include `-> &mut Self` returns.

---

Regardless, the `Spawning` structure looks good, and I'm OK with getting rid of the nicer-looking `spawn` method for the sake of consistency.

screeps-game-api/src/objects/mod.rs
```diff
@@ -289,6 +291,15 @@ pub unsafe trait CanStoreEnergy: StructureProperties {
     }
 }
 
+/// Used to specify which structures can use their stored energy for spawning creeps.
+///
+/// # Contract
+///
+/// The reference returned from `AsRef<Reference>::as_ref` must be have an
+/// `energy` and an `energyCapacity` properties. Also, that energy can be used
+/// by a spawner to create a new creep.
```
daboross: Small nit: this fact should already be implied by the thing implementing the contract for `CanStoreEnergy`. Would this structure be something like "... must be accepted by `energyStructures` option in `StructureSpawn.spawnCreep`"?

daboross:
Regarding

> At the same time, I made it so it is actually supported via a spawn_creep_with_options method.

I believe `SpawningOptions` should already be fully supported through the `.execute()` method. Is it currently broken?

screeps-game-api/src/objects/mod.rs
ASalvail: Aren't those two formulation equivalent? I can remove the part about `energy` and `energyCapacity` though.

ASalvail:
> I believe `SpawningOptions` should already be fully supported through the .execute() method. Is it currently broken?

Yes, I realized after the fact that this is what you had in mind afterwards. Without the small example, I was baffled as to why the implementation was done this way and I just assumed that it was due to being done at the beginning of the project.

screeps-game-api/src/objects/mod.rs
daboross: `spawnCreep` specifically checks that the structures in `energyStructures` are `StructureSpawn` or `StructureExtension`: https://github.com/screeps/engine/blob/9d9c6165a48efd31ee3cb87730fb5092a3b10b9f/src/game/structures.js#L1119 - it won't accept just any object with `energy` and `energyCapacity`.

`HasEnergy` on the other hand just requires that it wraps some JS object which has `energy` and `energyCapacity`, right? `StructureLab` is `HasEnergy` and it has `energy` and `energyCapacity` fields, but it wouldn't be valid to implement `HasEnergyForSpawn` for it since `spawnCreep` would not accept it.

daboross: oh, nevermind, you did have that in there. My apologies I completely missed the "also, that energy can be used by a spawner to create a new creep".

daboross:
Yeah, it was a bit haphazard and untested. Still, it was _somewhat_ thought through.

PR #87, opened 2018-10-14
Implements look[At][Area]()

### LINKED ISSUE #81: Look[At][Area] return values -- How to represent them
- [lookAt](https://docs.screeps.com/api/#Room.lookAt)
- [lookAtArea](https://docs.screeps.com/api/#Room.lookAtArea)
- [look](https://docs.screeps.com/api/#RoomPosition.look)

The return value of those methods is an array that looks like:
```js
[
    { type: 'creep', creep: {...} },
    { type: 'structure', structure: {...} },
    ...
    { type: 'terrain', terrain: 'swamp' }
]
```

To emulate this, we have a few options:
- Make a collection of struct to represent all possible `LOOK_*` possibilities. This would mean having an enum that would reflect `constants::Look` and handle the dispatching. This means parsing everything the function return in one pass or using `JsVec`. Return type: `Vec<LookResult>` / `JsVec<LookResult>`.
- Instead of making a new collection of structs, use what is already there along with dynamic dispatch. Return type: `Vec<Box<dyn LookConstant>>`.

My personal preference so far would be for the second option, but I'm open to suggestions.

ASalvail:
Closes #81 

Introduces two new structs: `LookResult` used for `look` and `lookAt` and `PositionedLookResult` for `lookAtArea`. Both have a manual `TryFrom<Value>` used for conversion.

### DIFF
--- a/screeps-game-api/src/objects/mod.rs
+++ b/screeps-game-api/src/objects/mod.rs
@@ -26,8 +26,9 @@ mod structure;
 pub use self::{
     impls::{
         AttackEvent, AttackType, BuildEvent, Event, EventType, ExitEvent, FindOptions,
-        HarvestEvent, HealEvent, HealType, LookResult, ObjectDestroyedEvent, Path, RepairEvent,
-        Reservation, ReserveControllerEvent, Sign, SpawnOptions, Step, UpgradeControllerEvent,
+        HarvestEvent, HealEvent, HealType, LookResult, ObjectDestroyedEvent, Path,
+        PositionedLookResult, RepairEvent, Reservation, ReserveControllerEvent, Sign, SpawnOptions,
+        Step, UpgradeControllerEvent,
     },
     structure::Structure,
 };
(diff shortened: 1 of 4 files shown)


screeps-game-api/src/objects/mod.rs
```diff
@@ -26,8 +26,8 @@ mod structure;
 pub use self::{
     impls::{
         AttackEvent, AttackType, BuildEvent, Event, EventType, ExitEvent, FindOptions,
-        HarvestEvent, HealEvent, HealType, ObjectDestroyedEvent, Path, RepairEvent, Reservation,
-        ReserveControllerEvent, Sign, SpawnOptions, Step, UpgradeControllerEvent,
+        HarvestEvent, HealEvent, HealType, LookResult, ObjectDestroyedEvent, Path, RepairEvent,
+        Reservation, ReserveControllerEvent, Sign, SpawnOptions, Step, UpgradeControllerEvent,
```
daboross: Minor nit: I think we want to re-export `PositionedLookResult` too?

daboross:
Darn it, I really shouldn't have selected the "dismiss reviews when new commits are pushed" option...

PR #92, opened 2018-10-26
Fix JavaScript code in StructureSpawn methods

daboross:
Two small fixes for `spawn_creep` and `spawn_creep_with_options` actually working.

I should have caught the first one when it was merged, but completely overlooked it. Really makes me wish we had CI/minimal testing of javascript interaction part of this, but that would be a huge pain to set up with the screeps runtime.

In the meantime, testing by running the latest version in a personal ai works! and it catches things like this.

See individual commits for descriptions of why changes were made.


Stargateur:
however https://github.com/daboross/screeps-in-rust-via-wasm/pull/92/commits/d77d87a88dae8baa8e017808e771cd4e51575a50 wasn't necessary, it's work fine in this form (at least for me).

PR #96, opened 2018-10-30
Add get<T>() and get_path<T>() for MemoryReference

### LINKED ISSUE #95: Get enum from memory reference
Hello,

I'd like to set and get an enum from the memory, but it seems like there are no getter for it. I found the MemoryReference::set<T: JsSerialize>() but there is no MemoryReference::get<T: JsDeserialize>(). Any idea how I could do that?

Thanks!

Stargateur:
Fixes #95


daboross:
Thanks for making this!

I think the travis failure is spurious, going to try rebuilding it.

PR #100, opened 2018-11-04
Add screeps-game-api CHANGELOG file

daboross:
The majority of these entries were copied from release notes stored in git tags. Some of the oldest ones were written up fresh from looking at diffs, and the unreleased section is from looking at PRs merged since the last release.


daboross:
I'm glad! Though it is a bit of work writing up each release...

PR #101, opened 2018-11-04
Clippy suggestions + formatting

daboross:
Most of this stuff should be pretty positive. Just following `clippy` suggestions that seem reasonable, mostly.

The one thing I'm not sure about use turning the `Density` variants like `Density::DensityLow` into `Density::Low`. It departs from the official constants, but it gets rid of stuttering, which is generally a good thing.

---

This isn't very high priority - I just needed to do a bit of cleanup on _something_, and screeps-game-api won't be worse off for it.

### DIFF
--- a/screeps-game-api/src/constants.rs
+++ b/screeps-game-api/src/constants.rs
@@ -668,10 +668,10 @@ enum_from_primitive! {
     #[repr(u32)]
     #[derive(Debug, PartialEq, Eq, Clone, Copy)]
     pub enum Density {
-        DensityLow = 1,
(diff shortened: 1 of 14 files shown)


daboross:
Thanks for looking over this!

With taking by value- yeah, things are copied. Clippy suggested it since copying a value 4 bytes and smaller is unconditionally faster than making and deferencing a reference to said value. https://rust-lang-nursery.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref is where it's documented at.

PR #105, opened 2018-12-03
Pass HasPosition by reference whenever possible

daboross:
Taking HasPosition things by value doesn't have much purpose, because the only method available on `HasPosition` takes `&self`. At least that's my reasoning?

We don't have any `impl<T> HasPosition for &T where T: HasPosition`. If we did, this would be unnecessary. I think taking `&T` in methods is simpler than this blanket impl, though, especially since we already have another blanket impl which would have to be removed if we added this one.


daboross:
:+1:

I _think_ it's worthwhile to do a quick non-breaking release before merging this, but that might be overdoing it.

PR #112, opened 2018-12-27
Removing allocation from PathFinder.search and Room.find calls

daboross:
This is slightly more unsafe code, but should still be sound. The principal is the same as transmuting `'a` to `'static` - it just affects the reference to the `Fn` as well now, not just the lifetime of the data within the box.

The actual change is that instead of allocating a space for the callback on the heap and storing a pointer to it, we just store a pointer to the stack location it's in within the pathfinder function. This change is made for both `Room.find` and `PathFinder.search`.

### DIFF
--- a/screeps-game-api/src/objects/impls/room.rs
+++ b/screeps-game-api/src/objects/impls/room.rs
@@ -32,7 +32,7 @@ simple_accessors! {
     // todo: visual
 }
 
-scoped_thread_local!(static COST_CALLBACK: Box<Fn(String, Reference) -> Option<Reference>>);
+scoped_thread_local!(static COST_CALLBACK: &'static Fn(String, Reference) -> Option<Reference>);
 
 impl Room {
     pub fn serialize_path(&self, path: &[Step]) -> String {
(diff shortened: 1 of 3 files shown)


screeps-game-api/src/objects/impls/room.rs
```diff
@@ -170,18 +170,17 @@ impl Room {
...
-        // of it, but it's still necessary because "some lifetime above the current scope but otherwise
-        // unknown" is not a valid lifetime to have PF_CALLBACK have.
-        let callback_lifetime_erased: Box<
-            Fn(String, Reference) -> Option<Reference> + 'static,
-        > = unsafe { mem::transmute(callback_type_erased) };
+        // now unified as &Fn
+        let callback_type_erased: &(Fn(String, Reference) -> Option<Reference> + 'a) =
+            &callback_boxed;
+
+        // Overwrite lifetime of box inside closure so it can be stuck in scoped_thread_local
```
ASalvail: Still refers to the `Box`

screeps-game-api/src/pathfinder.rs
```diff
@@ -360,15 +360,15 @@ where
     let callback_unboxed = move |input| raw_callback(input).inner;
 
     // Type erased and boxed callback: no longer a type specific to the closure passed in,
-    // now unified as Box<Fn>
-    let callback_type_erased: Box<Fn(String) -> Reference + 'a> = Box::new(callback_unboxed);
+    // now unified as &Fn
+    let callback_type_erased: &(Fn(String) -> Reference + 'a) = &callback_unboxed;
 
     // Overwrite lifetime of box inside closure so it can be stuck in scoped_thread_local storage:
```
ASalvail: Same with `Box`.

daboross: These should be fixed!

daboross:
Rebased on master for changelog conflicts. Maybe not including those in PRs would be a good idea in the future, just for the sake of not conflicting on editing the same file... But it's nice having it already there? Not sure.

I think I've fixed all the box comments.

PR #114, opened 2018-12-27
Create and implement `JsVec`

daboross:
This is split off of #73 for merging first. I'm not confident on how we want to implement `JsMap`, but `JsVec` seems good.

This is also rebased on master from #73.


daboross:
Alright!

I've rebased from master- if you'd consider it good I'll merge this.

ASalvail:
Looking good. My remaining questions are really for my education and already tracked in another PR.
Thanks @daboross !

PR #116, opened 2018-12-27
Switch from enum-primitive to num-derive

### LINKED ISSUE #72: Investigate migrating to `num-derive` for integer-based enums.
As mentioned in #57, `num-derive` might be a better choice going forward rather than `enum-primitives`. It should be investigated and maybe implemented.

daboross:
This does indeed simplify things!

Closes #72.


daboross:
It'll be interesting!

PR #122, opened 2019-01-10
Dyn trait structure accessors

### LINKED ISSUE #121: Conversions from Structure to trait objects
For usability, it'd be nice to have quick ways to go from `&Structure` to `Option<&dyn Attackable>`, or `Option<&dyn HasEnergyForSpawn`.

I'm thinking of using `&dyn Trait` here mostly because it's cheap (no allocation if it's just borrowing), and won't require much additional code (we just need one method on `Structure` per trait).

It's bad in the sense that it's very much manual code, and won't scale without more methods for each trait. I'm not sure if I can think of a particularly better option, though, mostly since most of our traits do not refer to actual JavaScript types. If they were JS types, we could use [`Reference::downcast`](https://docs.rs/stdweb/0.4.12/stdweb/struct.Reference.html#method.downcast), but things like `Attackable` and `HasEnergyForSpawn` just aren't.

The other disadvantage is that it'll require us to remove a few bounds from `RoomObjectProperties`. Specifically, we'll need to move the `Into<Reference>`, `TryFrom<...>`  and `ReferenceType` requirements to another disjoint trait, since they all require anything implementing them (and thus `RoomObjectProperties`) to have a concrete underlying `Sized` type. In other words, there can never be a method `<dyn Attackable>::try_from`, so as long as our traits all depend on `RoomObjectProperties`, it should not depend on those traits requiring `Sized`.

Third disadvantage is that this involves double-indirection doing method calls on the resulting trait objects. We'll be using a vtable holding a reference to the function specific to each structure, but the functions are implemented practically identically, and just call into JS. This might not be a big deal now, but I imagine it'd make this a bit slower.

daboross:
Closes #121 (see that issue).

### DIFF
--- a/screeps-game-api/src/objects/structure.rs
+++ b/screeps-game-api/src/objects/structure.rs
@@ -2,6 +2,7 @@ use stdweb::{InstanceOf, Reference, ReferenceType, Value};
 
 use {
     constants::StructureType,
+    objects::{Attackable, CanDecay, CanStoreEnergy, HasCooldown, HasEnergyForSpawn, HasStore},
     traits::{FromExpectedType, IntoExpectedType, TryFrom, TryInto},
     ConversionError,
 };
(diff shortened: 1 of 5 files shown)


screeps-game-api/src/objects/structure.rs
```diff
@@ -30,6 +31,126 @@ pub enum Structure {
...
+            self,
+            {
+                Container, Extension, Lab, Link, PowerSpawn, Spawn, Storage, Terminal, Tower
+            },
+            v => v
+        )
+    }
+
+    pub fn as_attackable(&self) -> Option<&dyn Attackable> {
+        match self {
```
ASalvail: Why didn't you use the macro here? I can't figure out.

ASalvail: A macro for two functions just makes the code less readable, don't bother.

PR #123, opened 2019-01-10
Fix Attackable impl on Structure for controllers

daboross:
This implements the fix "just return 0 for controllers for hits and hits_max". Note that `StructureController` still lacks an `Attackable` implementation- it's just on a `Structure` that this happens.

See https://github.com/daboross/screeps-in-rust-via-wasm/issues/118#issuecomment-452740686.


ASalvail:
Given #122, I'd remove `Attackable` from structure completely, without providing the modified version that return zeros.

PR #130, opened 2019-02-11
Correcting Command::Copy

mettke:
Changing operation from upload to copy


ASalvail:
Oops, we missed that one. Thanks for the contribution!

PR #134, opened 2019-02-20
Changing visibility of types required for api

mettke:
* LocalRoomNameParseError:
https://docs.rs/screeps-game-api/0.4.0/screeps/struct.LocalRoomName.html
* Bodypart, MoveToOptions
https://docs.rs/screeps-game-api/0.4.0/screeps/objects/struct.Creep.html
* PortalDestination
https://docs.rs/screeps-game-api/0.4.0/screeps/objects/struct.StructurePortal.html

### DIFF
--- a/screeps-game-api/src/objects/impls/mod.rs
+++ b/screeps-game-api/src/objects/impls/mod.rs
@@ -25,11 +25,13 @@ mod structure_tower;
 mod tombstone;
 
 pub use self::{
+    creep::{Bodypart, MoveToOptions},
     room::{
         AttackEvent, AttackType, BuildEvent, Event, EventType, ExitEvent, FindOptions,
         HarvestEvent, HealEvent, HealType, LookResult, ObjectDestroyedEvent, Path,
         PositionedLookResult, RepairEvent, ReserveControllerEvent, Step, UpgradeControllerEvent,
     },
     structure_controller::{Reservation, Sign},
+    structure_portal::PortalDestination,
     structure_spawn::SpawnOptions,
 };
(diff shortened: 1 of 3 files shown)


screeps-game-api/src/objects/impls/mod.rs
```diff
@@ -30,6 +30,8 @@ pub use self::{
         HarvestEvent, HealEvent, HealType, LookResult, ObjectDestroyedEvent, Path,
         PositionedLookResult, RepairEvent, ReserveControllerEvent, Step, UpgradeControllerEvent,
     },
+    creep::{Bodypart, MoveToOptions},
```
ASalvail: Those should be accessible through their own sub-modules, aren't they?

mettke:
Sry, I thought linking the docs would explain the problem sufficiently. Lets elaborate a bit.

In my experience with the rustdoc, if the documentation does not show a type it is not visible to the outside. This is backed up by the fact that `mod positions;` and `mod impls;` are both private modules. Thus, even thou the types themselves are marked as public, they will only be available to submodules and not a third party.

daboross:
Thanks!

I think this is definitely my fault for having a confusing module structure which allows for public-marked types which don't end up actually public.

ASalvail:
Sure! I'm waiting for the changes to be passed through `rustfmt` to approve.

mettke:
@ASalvail I see - sry guess I was a bit confused or something. Completely missed the point of clippy and rustfmt. Will do that this afternoon.

daboross:
Hi!

Sorry, I completely missed that last commit - had the notification for the comment but missed that the change had been made.

@ASalvail Hope it's alright if I just merge this - the formatting seems to be fixed up.

And thanks again @mettke!

ASalvail:
I also missed the commit, so I'm good. Thanks @daboross 
And thanks @mettke for this PR (and your patience!)

PR #135, opened 2019-02-23
Fix Room::find panics on the live environment

snorrwe:
Hey!

I encountered an issue where `Room::find` would panic when playing on the live environment.
it works fine in simulation, but on the live world I always encounter a panic.
Here's an example stack trace
```
[8:18:11 PM][shard3]Encountered a panic!
[8:18:11 PM][shard3]Panic error message: js_unwrap_ref at 97 in C:\Dev\screeps-in-rust-via-wasm\screeps-game-api\src\objects\impls\room.rs: Custom("reference is of a different type")
[8:18:11 PM][shard3]Panic location: src/libcore/result.rs:1009
[8:18:11 PM][shard3]caught exception: RuntimeError: unreachable
[8:18:11 PM][shard3]stack trace: RuntimeError: unreachable
    at rust_panic (wasm-function[613]:1)
    at _ZN3std9panicking20rust_panic_with_hook17h869ce4541d4e3554E (wasm-function[451]:295)
    at _ZN3std9panicking18continue_panic_fmt17h0ad726c0b188da91E (wasm-function[629]:75)
    at rust_begin_unwind (wasm-function[662]:3)
    at _ZN4core9panicking9panic_fmt17h4d67173bc68f6d5aE (wasm-function[438]:70)
    at _ZN4core6result13unwrap_failed17hfc6bc97b52ed5f63E (wasm-function[288]:130)
    at _ZN5xenos6creeps9harvester15attempt_harvest17h865640d46b5bd9d4E (wasm-function[119]:4503)
    at _ZN5xenos6creeps9harvester3run28_$u7b$$u7b$closure$u7d$$u7d$17h300928efa798c577E (wasm-function[321]:10)
    at _ZN70_$LT$xenos..bt..Control$LT$$u27$a$GT$$u20$as$u20$xenos..bt..BtNode$GT$4tick17h631da085a125a2dbE (wasm-function[115]:1096)
    at _ZN5xenos6creeps9harvester3run28_$u7b$$u7b$closure$u7d$$u7d$17h80e90cff34d67169E (wasm-function[280]:7)
```

Calling `Object.values` on the JavaScript results before returning seems to fix this problem.

### DIFF
--- a/screeps-game-api/src/objects/impls/room.rs
+++ b/screeps-game-api/src/objects/impls/room.rs
@@ -94,7 +94,7 @@ impl Room {
     where
         T: FindConstant,
     {
-        js_unwrap_ref!(@{self.as_ref()}.find(@{ty.find_code()}))
+        js_unwrap_ref!(Object.values(@{self.as_ref()}.find(@{ty.find_code()})))
     }
 
     pub fn find_exit_to(&self, room: &Room) -> Result<Exit, ReturnCode> {


ASalvail:
Given https://docs.screeps.com/api/#Room.find the find method should return an `Array`. Your workaround seems to indicate that this is not the case, that it returns an object.
Calling `Objects.values` would mean we are discarding the keys of the values, wouldn't it? In that case, what is it that we are discarding?

snorrwe:
I whipped up a small test:

```
    js!{
        const room = @{room};
        const sources = room.find(FIND_SOURCES);
        console.log("|", sources, "|", Object.keys(sources), "|", Object.values(sources));
    }
```

And got the following result:

```
[8:52:59 PM][shard3]| [source #5bbcaaf29099fc012e6327cd],[source #5bbcaaf29099fc012e6327cf] | 0,1 | [source #5bbcaaf29099fc012e6327cd],[source #5bbcaaf29099fc012e6327cf]
```

It seems like we're discarding the indices of the list. On the JavaScript side it works as an Array, but stdweb can not parse it as such.
I believe that this behaviour is a bug in the Screeps implementation.

daboross:
@snorrwe Do you perhaps have the [IVM runtime](https://blog.screeps.com/2018/03/changelog-2018-03-05/) disabled on screeps? We currently only support IVM (though I'm not sure if that's documented anywhere...)

In the non-IVM runtime, the javascript virtual machine creates new types for use inside the VM which are identical to the ones outside, but are technically different. In particular, player code has access to different "Object" and "Array" types than code outside. It _usually_ works alright, but this means that `instanceof Array` doesn't work for things returned from game calls, and prototype modifications on Array don't exist. We used to have a workaround for this, but it was removed in https://github.com/daboross/screeps-in-rust-via-wasm/commit/ae11e89b8ecb90c570c0d415d2a582c78f1e102a shortly after the IVM runtime was officially released.

---

Could you check and confirm which runtime you're using and whether this still occurs with the IVM?

snorrwe:
@daboross You're right, I was using the legacy VM. The panic does not occur with IVM enabled. I'm a returning player, I had no idea this setting existed. I think this might be worth a mention in the documentation, but since it's deprecated I don't think this workaround should be included in this repo. I'll close the PR.

daboross:
Definitely should be referenced in docs! I've opened https://github.com/daboross/screeps-in-rust-via-wasm/issues/136 so this isn't forgotten.

PR #137, opened 2019-03-24
Add support for using Screeps auth tokens instead of username/password

npfund:
Add support for optionally supplying a Screeps authentication token for upload purposes.

### DIFF
--- a/cargo-screeps/src/upload.rs
+++ b/cargo-screeps/src/upload.rs
@@ -2,7 +2,7 @@ use std::{collections::HashMap, fs, io::Read, path::Path};
 
 use {base64, failure, reqwest, serde_json};
 
-use config::Configuration;
+use config::{Configuration, UploadConfiguration};
 
 pub fn upload(root: &Path, config: &Configuration) -> Result<(), failure::Error> {
     let upload_config = config.upload.as_ref().ok_or_else(|| {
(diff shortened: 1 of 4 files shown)


cargo-screeps/src/upload.rs
```diff
@@ -94,3 +92,22 @@ pub fn upload(root: &Path, config: &Configuration) -> Result<(), failure::Error>
...
+            "X-Token",
+            upload_config.auth_token.as_ref().unwrap().as_str(),
+        ))
+    } else if upload_config.username.is_some() && upload_config.password.is_some() {
+        Ok(request.basic_auth(
+            upload_config.username.as_ref().unwrap(),
+            upload_config.password.as_ref(),
+        ))
+    } else {
+        bail!("could not authorize request. must set auth_token or username/password in [upload] config");
```
daboross: Ideally `UploadConfiguration` should always represent a valid configuration, and we shouldn't have to bail out later because of invalid configuration options.

Do you think we could bail out here instead?

 https://github.com/daboross/screeps-in-rust-via-wasm/blob/430c097ea9d8340dfa2aa6d2fe26c475af949c85/cargo-screeps/src/config.rs#L110-L134

We could then represent the authentication with something like

```rust
pub struct Authentication {
    Token(String),
    UserPass { username: String, password: String },
}
```

Or just leave it as is, and make this bail an `unreachable!()` panic.

npfund: Makes sense! I'll see what I can do.

npfund:
Looks like nightly failed to install for the mac build?

daboross:
Yeah, looks pretty spurious. I've restarted the build.

Looks good though! :+1:

daboross:
Just realized that travis never seemed to update github on this, and I never merged this. Apologies for that!

Merging this now. Thanks again for the work on this!

PR #142, opened 2019-06-15
add safeguard to compile only when sufficient CPU available

reggaemuffin:


daboross:
Looks good to me! I left things like this out initially thinking that it'd help promote [making custom initialization headers per project](https://github.com/daboross/screeps-in-rust-via-wasm/blob/master/cargo-screeps/docs/initialization-header.md), but having this one check in can't hurt.

cargo-screeps/resources/default_initialization_header.js
```diff
@@ -2,6 +2,12 @@
 let wasm_module = null;
 
 function wasm_initialize() {
+    
+    if(Game.cpu.bucket < 500){
```
daboross: We could have some spacing around the parens here, but that can always be changed afterwards.

reggaemuffin:
Yeah I already checked the cargo-web prepend-js directives but they append outside that wasm-init function, so they are not that useful. TBH the message was kinda uncreative 😆

cargo-screeps/resources/default_initialization_header.js
reggaemuffin: you can also edit the pr directly from the web interface if you want to fix style issues :)

daboross:
Fair - it's built that way! The `wasm_initialize` is meant to have the most control, so that someone overriding `initialization-header` can change as many things as they like like this per-project.

I'll wait a bit to see if @ASalvail has any objections, but then we can just merge it.

ASalvail:
Since this is somewhat needed due to the use of rust, I say it has its place in this repo.

Thanks @reggaemuffin !

Also, it might be nice to start another repo of tools and goodies we can build on top of this interface to make life easier. Not AI per se, but quality of life that goes beyond simply interfacing. Food for thought.

PR #145, opened 2019-06-19
Implemented Into() for numeric tuple types on RoomPosition

Dessix:
This adds `Into` implementations for `RoomPosition` for pairs of `u8` through `u64` and `i8` through `i64`.

This also adds `coords() -> (u32, u32)` and `coords_signed() -> (i32, i32)` to `impl RoomPosition`.


daboross:
Since this is pretty minimal, and can be useful, just going to merge. Can fix up before release if necessary?

PR #148, opened 2019-06-21
Implement Hash for constants

daboross:
Needed this on one constant for putting it into a `HashMap` as a key. I don't think it adds too much bloat?


daboross:
Glad to be fixing this up!

I've actually been working on my AI recently, and just using the codebase more has been a good way to find things it can't do.

PR #149, opened 2019-06-22
Added naive string equality for LocalRoomName

Dessix:
This allows comparison between `LocalRoomName` and `&str`, `String`, or `&String`.
Internally, this simply formats the `LocalRoomName` and compares, but this can be improved with parser-like behaviour if profiling deems it necessary.

### DIFF
--- a/screeps-game-api/src/positions.rs
+++ b/screeps-game-api/src/positions.rs
@@ -205,6 +205,62 @@ impl<'a> fmt::Display for LocalRoomNameParseError<'a> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


daboross:
I think I'd be more happy using `parse_or_cheap_failure` than formatting / allocating a string, but I guess it can be optimized later.

@ASalvail any thoughts on this?

ASalvail:
Since it can't be done by them locally, it's not a bad thing. I don't suppose there's going to be that many string comparisons happening, so the string allocation might not be so bad.

We could add an issue to track switching to a cheaper alternative.

PR #157, opened 2019-07-02
Fix flag creation result interpretation.

daboross:
I've tested this locally. Would be nice if we had unit tests set up with the screeps server, but that's a whole deal.

Fixes #115.

I'm certain `Flag` isn't _really_ where `interpret_creation_ret_value` should go, but it does relate to flags. Might be worth moving small code reuse functions like this to a util(s) module in the future?


daboross:
Sounds reasonable to me! We can fix it up if/when we find a better way to represent things.

PR #160, opened 2019-07-07
Add a method to stop tracking a RawMemory segment using `drop_segment`.

ASalvail:
It turns out that the Screeps driver doesn't check which segments are written at the end of a tick, so long as there are only 10 segments that are written to.

This leads to a bot being able to "touch" 20 segments in a tick: 
- 10 to be read at the beginning of the tick (from the `setActiveSegments`)
- 10 to be written at the end of the tick (in `RawMemory.segments`)

However, no checks is done to verify that the written segments are part of the active segments.
Thus, by adding a way to `drop_segment` (this PR) (deleting it from `RawMemory.segments`), it can make space to write to a different segment.

Note: deleting `RawMemory.segments[k]` is local only, it doesn't touch the actual data in the segment.

### DIFF
--- a/screeps-game-api/src/raw_memory.rs
+++ b/screeps-game-api/src/raw_memory.rs
@@ -33,6 +33,12 @@ pub fn set_segment(id: u32, data: &str) {
     }
 }
 
+pub fn drop_segment(id: u32) {
+    js! { @(no_return)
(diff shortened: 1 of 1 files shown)


daboross:
This is a pretty awesome thing, though! I had no idea we could manipulate segments written like this.

PR #162, opened 2019-07-10
Make many StructureController properties Option<u32>.

daboross:
These properties seem to be guaranteed undefined if they're not applicable.

Found this trying to use them, then found the complete list of sometimes-undefined properties by checking accessor code at https://github.com/screeps/engine/blob/a4e431c9ee33dcc55c04f3adc479145c210c73fd/src/game/structures.js#L263-L268.


daboross:
Yeah I mean... who knows? `undefined` could be thought of as more semantically accurate, like the controller with level 0 doesn't have 0 progress, it has no progress? or like safe mode won't expire in 0 ticks, it isn't present.

but that's a bit of an arbitrary difference I guess. Eh

Issue #88, opened 2018-10-15
Update constants


ASalvail:
- [ ] Update the constants (check whether any is missing.
- [ ] Uniformize the derived traits (I think @daboross already did a pass on that) 
- [ ] Uniformize the implementation of `From` and `TryFrom` for all structs and enums defined.
- [ ] Split the module into submodules to make it easier to access.

PR #164, opened 2019-07-13
Constants overhaul

### LINKED ISSUE #58: Constant SIGN_PLANNED_AREA
Added as of https://blog.screeps.com/2018/09/changelog-2018-09-22/.

See https://github.com/screeps/common/blob/560c3937e9f9b279eae5ef939f07b2e618452e43/lib/constants.js#L699.

daboross:
This implements the ability to run `look` and `find` on power creeps, adds some power creep related constants, and adds other misc. constant updates.

This does add a `PowerCreep` type, but it doesn't add any implementation to it beyond it being a `RoomObject`. This should be enough to at least notice power creeps existing, but not really enough to do anything with one's own power creeps.

Fixes #58.


screeps-game-api/javascript/utils.js
```diff
@@ -89,7 +90,7 @@ function __structure_type_str_to_num(str) {
         case STRUCTURE_TERMINAL: return 16;
         case STRUCTURE_CONTAINER: return 17;
         case STRUCTURE_NUKER: return 18;
-        default: throw new Error("unknown resource type " + str);
+        default: throw new Error("unknown structure type " + str);
```
ASalvail: Haha, good catch!

screeps-game-api/src/constants.rs
```diff
@@ -979,3 +995,60 @@ impl TryFrom<Value> for ResourceType {
...
+pub const POWER_LEVEL_POW: u32 = 2;
+pub const POWER_CREEP_SPAWN_COOLDOWN: u32 = 8 * 3600 * 1000;
+pub const POWER_CREEP_DELETE_COOLDOWN: u32 = 24 * 3600 * 1000;
+pub const POWER_CREEP_MAX_LEVEL: u32 = 25;
+pub const POWER_CREEP_LIFE_TIME: u32 = 5000;
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "constants-serde", derive(Serialize, Deserialize))]
+pub enum PowerClass {
+    #[cfg_attr(feature = "constants-serde", serde(rename = "operator"))]
```
ASalvail: You just removed those higher up. Can this one also be changed to `#[serde(rename(...)]`?

screeps-game-api/src/constants.rs
```diff
@@ -979,3 +995,60 @@ impl TryFrom<Value> for ResourceType {
...
+    RegenSource = 13,
+    RegenMineral = 14,
+    DisruptTerminal = 15,
+    OperatePower = 16,
+    Fortify = 17,
+    OperateController = 18,
+    OperateFactory = 19,
+}
+
+// TODO: POWER_INFO
```
ASalvail: Wasn't there anything about tunnels also? I'm uncomfortable bumping the date at the top of the file without catching everything that was updated in the constants.
Also, how hard could it be to make a macro that uses the `js` file that describes the constants directly to generate this file?

screeps-game-api/src/lib.rs
```diff
@@ -29,6 +29,7 @@ extern crate serde;
 #[macro_use]
 extern crate serde_derive;
 extern crate serde_json;
+extern crate serde_repr;
```
ASalvail: You'll have to rebase off of the 2018 rustification 😞

screeps-game-api/src/constants.rs
daboross: Hm - I mean we could! I removed the ones above because we used them within the codebase, while these are still feature gated so we don't generate extra code. It is somewhat arbitrary that serde implementations are feature gated this way but nothing else is, though.

Maybe we should remove these gates and just have everything implement it all the time!

screeps-game-api/src/constants.rs
daboross: I think the only tunnel thing was `CONSTRUCTION_COST_ROAD_WALL_RATIO`, which I added above.

But I agree with the concerns here - I'll either implement the rest of them or make a summary TODO and TODO issue.

daboross: With automated stuff, I bet we could do something for all the "plain" constants which are just numbers. But I think it might be hard translating to enums like ResourceType, PowerType, and such. I mean I haven't even figured out how we want to translate reactions nor power info - I _think_ functions which match on arguments and return the result would be the best way, but that might not be true.

daboross:
Rebased on master - going to add more constants though to see how close I can get to completion.

screeps-game-api/src/constants.rs
ASalvail: Do you remember why we wanted to feature gate them? While we might save some space, this is something that would trip newcomers.

screeps-game-api/src/constants.rs
ASalvail: Maybe later...

screeps-game-api/src/constants.rs
daboross: I don't really - I think I just personally needed them and didn't want to bloat the binary?

I'll remove the feature gate - they're useful enough.

daboross:
Rebased after #166 and did a lot more constant management.

Added some breaking changes which may or may not want to be part of this PR?

screeps-game-api/src/constants.rs
```diff
@@ -16,6 +16,7 @@
 
 use log::error;
 use num_derive::FromPrimitive;
+use num_traits::FromPrimitive;
```
ASalvail: Isn't that shadowing the previous import?

screeps-game-api/src/constants.rs
```diff
@@ -451,30 +447,12 @@ impl Part {
...
-}
-
 impl TryFrom<Value> for Part {
     type Error = ConversionError;
     fn try_from(v: Value) -> Result<Self, Self::Error> {
         let x: u32 = v.try_into()?;
-        Ok(Self::try_from(x)
-            .unwrap_or_else(|()| panic!("JavaScript gave unknown part constant {}", x)))
+        Ok(Self::from_u32(x)
+            .unwrap_or_else(|| panic!("JavaScript gave unknown part constant {}", x)))
```
ASalvail: Shouldn't you return an `Err`?

screeps-game-api/src/constants.rs
```diff
@@ -655,28 +633,7 @@ impl TryFrom<Value> for StructureType {
...
-            12 => StructureType::PowerBank,
-            13 => StructureType::PowerSpawn,
-            14 => StructureType::Extractor,
-            15 => StructureType::Lab,
-            16 => StructureType::Terminal,
-            17 => StructureType::Container,
-            18 => StructureType::Nuker,
-            _ => panic!("unknown structure type integer {}", x),
-        })
+        Ok(Self::from_u32(x).unwrap_or_else(|| panic!("unknown structure type integer {}", x)))
```
ASalvail: Same with `Err`.

screeps-game-api/src/constants.rs
```diff
@@ -1115,52 +1070,8 @@ impl TryFrom<Value> for ResourceType {
...
-            38 => ResourceType::CatalyzedLemergiumAcid,
-            39 => ResourceType::CatalyzedLemergiumAlkalide,
-            40 => ResourceType::CatalyzedZynthiumAcid,
-            41 => ResourceType::CatalyzedZynthiumAlkalide,
-            42 => ResourceType::CatalyzedGhodiumAcid,
-            43 => ResourceType::CatalyzedGhodiumAlkalide,
-            _ => panic!("unknown resource type integer {}", x),
-        })
+
+        Ok(Self::from_u32(x).unwrap_or_else(|| panic!("unknown resource type integer {}", x)))
```
ASalvail: `Err`

screeps-game-api/src/constants.rs
```diff
@@ -1217,3 +1129,13 @@ pub enum PowerType {
...
     OperateFactory = 19,
 }
+
+impl TryFrom<Value> for PowerType {
+    type Error = ConversionError;
+
+    fn try_from(v: Value) -> Result<Self, Self::Error> {
+        let x: u32 = v.try_into()?;
+
+        Ok(Self::from_u32(x).unwrap_or_else(|| panic!("unknown power type integer {}", x)))
```
ASalvail: `Err`

screeps-game-api/src/constants.rs
```diff
@@ -14,13 +14,12 @@
...
 //! [the game constants]: https://github.com/screeps/common/blob/master/lib/constants.js
 
-use log::error;
 use num_derive::FromPrimitive;
 use num_traits::FromPrimitive;
 #[allow(unused_imports)]
 use serde::{Deserialize, Serialize};
 use serde_repr::{Deserialize_repr, Serialize_repr};
-use stdweb::{Number, Reference, Value};
+use stdweb::{Number, Reference, Value, __js_deserializable_serde_boilerplate, js_deserializable};
```
ASalvail: if you want to avoid weird macro imports all over the place, feel free to use `crate::macros::*` which exposes those macros.

screeps-game-api/src/constants.rs
```diff
@@ -233,17 +217,7 @@ pub enum Direction {
...
-
-    fn try_from(v: Value) -> Result<Self, Self::Error> {
-        let as_num = u32::try_from(v)?;
-
-        Ok(Self::from_u32(as_num).unwrap_or_else(|| {
-            panic!("encountered a direction code we don't know: {}", as_num);
-        }))
-    }
-}
+js_deserializable!(Direction);
```
ASalvail: I thought we had agreed not to use `js_deserializable!` since it was too costly?

screeps-game-api/src/constants.rs
```diff
@@ -513,7 +517,7 @@ pub const STORAGE_HITS: u32 = 10_000;
 
 /// Translates `STRUCTURE_*` constants.
 #[repr(u32)]
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, FromPrimitive, Serialize, Deserialize)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
```
ASalvail: I'm not sure I understand the difference between `Exit` which uses `FromPrimitive` and `StructureType` which doesn't. They're both integers. Unless you want to be pedantic and say that `u32` are natural numbers and not integers 😛

screeps-game-api/src/constants.rs
```diff
@@ -896,47 +896,82 @@ pub enum ResourceType {
...
-    pub fn reaction_time(self) -> u32 {
+    /// Translates the `REACTION_TIME` constant.
+    pub fn reaction_time(self) -> Option<u32> {
         use crate::ResourceType::*;
-        match self {
+        let time = match self {
             Energy | Power | Hydrogen | Oxygen | Utrium | Lemergium | Keanium | Zynthium
-            | Catalyst | Ghodium | Ops => 0,
+            | Catalyst | Ops => return None,
+            // OH: 20,
```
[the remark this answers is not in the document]
daboross: Yeah! That and to double check I actually have the right values and to make it easier to compare to future `constants.js` files.

screeps-game-api/src/constants.rs
daboross: One's the derive and the other's the actual trait for usage - I think both are necessary?

screeps-game-api/src/constants.rs
daboross: I think it should be better with `stdweb` 0.4.9+! Deserializing now no longer uses json parsing, and instead implements `TryFrom<Value>` by traversing the value with serde, calling into javascript to get any references.

I think it should be fairly equivalent to a manual implementation of `TryFrom` now, especially with these integer constants which should just be represented by `Value::Number`.

screeps-game-api/src/constants.rs
daboross: Main difference is that in screeps `StructureType` is actually represented by strings like `"spawn"`, `"extension"`, `"road"`, etc. and the numbering code we use is entirely internal to `screeps-game-api`. I mean it uses `__structure_type_num_to_str` from `javascript/utils.js` to translate integers to the actual constant names.

I was thinking it'd be nice to have `FromPrimitive` for screeps constants which actually are integer constants, like Exit, Direction, ReturnCode, Find*, Color since users might actually encounter those? But it's mostly contrived reasoning to find some way to make it reasonable to have `FromPrimitive` on `ReturnCode` in particular since we use that.

daboross:
Thanks for the review! I have a lot to fix up and add comments for.

> I'm concerned with the very uneven treatment of TryFrom implementations

With this, I think it should be fairly consistent in that everything implements `TryFrom<Value>`, but I agree that the "extra" impls are kind of scattered around... I left in the ones for `ReturnCode` and `Exit` mostly because I didn't want to touch more code outside of the constants file, but I'll go in and remove those now.

screeps-game-api/src/constants.rs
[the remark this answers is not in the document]
ASalvail: Ah, you copy pasted the values, commented them then inserted the variants. It makes more sense now.

screeps-game-api/src/constants.rs
ASalvail: TIL that macros and items are not in the same namespace. Interesting!

screeps-game-api/src/constants.rs
ASalvail: Oh, that's great news, I didn't know!

ASalvail:
By uneven treatment, I mostly referred to the `try_from` implementations that `panic!` rather than return an error. I get that it should probably never happen, but it also sounds like the exact reason we're doing `try_from` rather than `from`.

ASalvail:
Fix #58

ASalvail:
Addresses some of #88

screeps-game-api/src/constants.rs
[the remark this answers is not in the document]
daboross: Yep!

screeps-game-api/javascript/utils.js
```diff
@@ -188,6 +190,7 @@ function __resource_type_str_to_num(str) {
         case RESOURCE_CATALYZED_ZYNTHIUM_ALKALIDE: return 41;
         case RESOURCE_CATALYZED_GHODIUM_ACID: return 42;
         case RESOURCE_CATALYZED_GHODIUM_ALKALIDE: return 43;
+        case RESOURCE_OPS: return 44;
```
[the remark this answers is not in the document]
daboross: I think it has something to do with power creeps? There's a power named `PWR_GENERATE_OPS` which seems like it could be related.

I found the constant `RESOURCE_OPS` in the constants file at https://github.com/screeps/common/blob/068343753adf090fd1991944d2292be9e236b7dd/lib/constants.js#L406, and it is counted as part of "all resources" like the other ones at https://github.com/screeps/common/blob/068343753adf090fd1991944d2292be9e236b7dd/lib/constants.js#L989.

screeps-game-api/src/constants.rs
daboross: it is! I should have said something though rather than just barging ahead.

screeps-game-api/src/constants.rs
```diff
@@ -732,6 +799,33 @@ impl TryFrom<Value> for Density {
...
+            Density::Moderate => 35_000,
+            Density::High => 70_000,
+            Density::Ultra => 100_000,
+        }
+    }
+
+    /// Translates the `MINERAL_DENSITY_PROBABILITY` constant.
+    ///
+    /// All values are between 0 and 1, but the total is roughly `2.5` so these
+    /// aren't percentages.
```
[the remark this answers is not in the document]
daboross: Ah - looks like this is right!

https://github.com/screeps/engine/blob/a4e431c9ee33dcc55c04f3adc479145c210c73fd/src/processor/intents/minerals/tick.js#L24-L29

daboross:
Re-rebased on master, and fixed up!

In addition to addressing the comments made here, it also includes some new `find` module improvements, and a small `Density` improvement.

As a note, I think there's _something_ that this PR breaks, since I've been running my AI using it and it's been panicking on a find call somewhere. I'm not sure what exactly is broken, but once I find it I'll make a note of it here.

screeps-game-api/src/constants.rs
```diff
@@ -199,6 +209,13 @@ pub mod find {
         MINERALS, 116, Mineral;
         NUKES, 117, Nuke;
         TOMBSTONES, 118, Tombstone;
+        POWER_CREEPS, 119, PowerCreep;
+        MY_POWER_CREEPS, 120, PowerCreep;
+        HOSTILE_POWER_CREEPS, 121, PowerCreep;
+        EXIT_TOP, Exit::Top as i32, RoomPosition;
+        EXIT_RIGHT, Exit::Right as i32, RoomPosition;
+        EXIT_BOTTOM, Exit::Bottom as i32, RoomPosition;
+        EXIT, Exit::All as i32, RoomPosition;
```
daboross: These should help with the situation we last talked about in the slack chat!

I might be putting too much in this PR. But what else is an "overhaul" for?

screeps-game-api/src/constants.rs
```diff
@@ -656,7 +681,31 @@ pub fn controller_levels(current_rcl: u32) -> u32 {
     }
 }
 
-// TODO: controller_*
+/// Translates the `CONTROLLER_DOWNGRADE` constant.
+///
+/// Accepts levels 1-7.
```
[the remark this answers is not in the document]
ASalvail: This is the only thing that wasn't fixed that I can see.

screeps-game-api/src/constants.rs
```diff
@@ -172,6 +211,10 @@ pub mod find {
         POWER_CREEPS, 119, PowerCreep;
         MY_POWER_CREEPS, 120, PowerCreep;
         HOSTILE_POWER_CREEPS, 121, PowerCreep;
+        EXIT_TOP, Exit::Top as i32, RoomPosition;
+        EXIT_RIGHT, Exit::Right as i32, RoomPosition;
```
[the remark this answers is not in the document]
daboross: thank you

screeps-game-api/src/constants.rs
```diff
@@ -59,54 +61,91 @@ impl ReturnCode {
...
-    fn find_code(&self) -> i32 {
-        *self as i32
+    /// Useful for finding any [`RoomObject`][crate::objects::RoomObject] with
+    /// a dynamically-chosen find constant.
+    ///
+    /// If you know ahead of time what constant you'll use, then the
+    /// all-upper-case constants in [this module](crate::constants::find) will
+    /// be more helpful, and won't require casting the result types.
+    ///
+    /// If you do use this constant, I recommend just importing the `find`
```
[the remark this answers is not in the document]
daboross: That sounds better, yeah. I wanted to write something "softer" than that, but since code really does look weird without `find::` beforehand "should" does seem good.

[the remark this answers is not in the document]
daboross: In other places I seem to use "it's recommended", which is slightly better. But _should_ in this case seems right, still.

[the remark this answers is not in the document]
ASalvail: _You may want_ ?

screeps-game-api/javascript/utils.js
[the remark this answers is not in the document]
daboross: It does seem to be used in engine: at https://github.com/screeps/engine/blob/a4e431c9ee33dcc55c04f3adc479145c210c73fd/src/processor/intents/power-creeps/usePower.js#L58.

screeps-game-api/src/constants.rs
```diff
@@ -658,7 +659,7 @@ pub const CONSTRUCTION_COST_ROAD_WALL_RATIO: u32 = 150;
 
 /// Translates the `CONTROLLER_LEVELS` constant.
 ///
-/// Accepts levels 1-7.
+/// Accepts levels 1-8.
```
ASalvail: Argh, no, this one is actually 1-7 😆

daboross: :man_facepalming:

daboross:
While I'm super excited for this, there is one problem: I assumed that `#[derive(Deserialize)]` in combination with serde-json and/or stdweb's Value deserializer implementation, enum variants could be deserialized from both integers and strings.

For these formats, when deserializing an enum, only string names for the variants are supported.

daboross:
Required adding a bit more code, but decoding all the constants which are encoded as integers but are actually strings in game should work now.

screeps-game-api/src/constants.rs
```diff
@@ -448,8 +458,16 @@ pub mod look {
     }
 }
 
+/// Creep part types.
+///
+/// *Note:* This constant's `TryFrom<Value>`, `Serialize` and `Deserialize`
+/// implementations only operate on made-up integer constants. If you're ever
+/// using these impls manually, you need to use the `__part_num_to_str` and
+/// `__part_str_to_num` JavaScript functions to convert.
```
ASalvail: hopefully you can copy paste the previous example...
Or maybe make one at the module level and refer to it here?

screeps-game-api/src/constants.rs
```diff
@@ -361,8 +361,8 @@ js_deserializable!(Color);
 ///
 /// *Note:* This constant's `TryFrom<Value>` and `Deserialize` implementations
 /// _only work with the integer constants_. If you're ever consuming strings
-/// such as `"plain"`, `"swamp"`, `"wall"`, you must convert those manually into
-/// `0`, `1` and `2` respectively.
+/// such as `"plain"`, `"swamp"`, `"wall"`, you can use the
```
ASalvail: Only in JS though. Might be worthwhile to make a `TryFrom<String>` equivalent?

screeps-game-api/src/constants.rs
```diff
@@ -390,39 +394,53 @@ js_deserializable!(Terrain);
 /// implementations only operate on made-up integer constants. If you're ever
 /// using these impls manually, use the `__look_num_to_str` and
 /// `__look_str_to_num` JavaScript functions to convert.
+///
+/// To convert from a string representation directly, use
```
ASalvail: As mentioned elsewhere, maybe specify that the best method depends on where the value lives?
Actually, it might be beneficial (in another PR) to add a brief explanation of the memory model in the crate's doc. It would avoid some confusion and explain design decisions. Thoughts?

screeps-game-api/src/constants.rs
```diff
@@ -390,39 +394,53 @@ js_deserializable!(Terrain);
...
     PowerCreeps = 11,
 }
 
 js_deserializable!(Look);
 
+impl Look {
+    /// Helper function for deserializing from a string rather than a fake
+    /// integer value.
+    pub fn deserialize_from_str<'de, D: Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
+        let s: Cow<'de, str> = Cow::deserialize(d)?;
```
ASalvail: I really don't get the usage of the `Cow`? Serde doesn't allocate memory for strings? Is it because it may still refers to the initial "text" that is deserialized?

daboross: `serde` won't do any allocations, but the deserializer implementations may or may not need to. `Cow` is just a way of encoding that this code is OK with taking either an allocated owned string, or a borrowed one.

For instance, when using `serde_json::from_str` to deserialize, strings without escape patterns can be deserialized as borrowing from the original string passed in. But in `serde_json::from_value`, the ownership of strings is passed from the value into the deserialized data, so we need to be able to accept an owned string too.

It'd function if this code just used `s: String`, but then it would also allocate slightly more than necessary in some situations.

screeps-game-api/src/constants.rs
daboross: Module-level doc sounds like a good idea.

screeps-game-api/src/constants.rs
daboross: Added! I gave it a `FromStr` implementation like some of the other constants.

screeps-game-api/src/constants.rs
```diff
@@ -380,8 +385,13 @@ js_deserializable!(Terrain);
 /// manually including JS code.
 ///
 /// To use in JS: `__look_num_to_str(@{look as u32})` function
+///
+/// *Note:* This constant's `TryFrom<Value>`, `Serialize` and `Deserialize`
+/// implementations only operate on made-up integer constants. If you're ever
+/// using these impls manually, you need to use the `__look_num_to_str` and
+/// `__look_str_to_num` JavaScript functions to convert.
```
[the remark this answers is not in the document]
daboross: Added an example to module-level doc.

screeps-game-api/src/constants.rs
daboross: Adding documentation on the memory model in general sounds like a good idea!

For this particular PR, I've mentioned the two different contexts where one could use the two different conversion methods in the longer documentation for the module.

PR #166, opened 2019-07-14
Use serde_repr crate!

daboross:
The https://github.com/dtolnay/serde-repr has existed for a bit now! It replaced the custom macro on https://serde.rs/enum-number.html, and this replaces usages of enum_number in this crate.

I think the main benefit should be better IDE support, since they generally read derives better than macros.


daboross:
I think we technically could, but it'd probably be more efficient to keep both? Since `serde_repr` still needs a serde deserializer running and `num_derive::FromPrimitive` only requires taking in a number.

daboross:
Force pushing a rebased version to fix merge conflicts.

daboross:
Merging this.

PR #168, opened 2019-07-14
Add some Display implementations for local values

daboross:
This just adds a few things I want to exist! Not necessarily the most consistent but could be useful.

### DIFF
--- a/screeps-game-api/src/constants.rs
+++ b/screeps-game-api/src/constants.rs
@@ -4,6 +4,7 @@
 //! <https://github.com/screeps/common/commits/master/lib/constants.js>.
 //!
 //! [the game constants]: https://github.com/screeps/common/blob/master/lib/constants.js
+use std::fmt;
 
 use log::error;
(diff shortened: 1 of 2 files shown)


screeps-game-api/src/constants.rs
```diff
@@ -261,6 +262,22 @@ impl ::std::ops::Neg for Direction {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let ch = match *self {
+            Direction::Top => "↑",
+            Direction::TopRight => "↗",
+            Direction::Right => "→",
+            Direction::BottomRight => "↘",
+            Direction::Bottom => "↓",
+            Direction::BottomLeft => "↙",
+            Direction::Left => "←",
+            Direction::TopLeft => "↖",
```
ASalvail: Neat!

daboross: Thanks! I thought it would be nice.

PR #171, opened 2019-07-16
Improve LocalRoomPosition's Serialize/Deserialize impls and add js_serializable! call

### LINKED ISSUE #159: Correctly using get<T> to get RoomPosition from creep memory.
So, I am trying to use the set<T> and get<T> methods to store and retrieve a screeps::RoomPosition to memory, but I'm not entirely sure how to correctly use this.

```
match creep.memory().get<screeps::RoomPosition>("target")
{
    Ok(target) => move_to(target),
    Err(_) => warn!("Something went wrong trying to get target from creep")
}
```
Is this the correct way to go use this method?

_Originally posted by @WooterTheTroubleshooter in https://github.com/daboross/screeps-in-rust-via-wasm/issues/95#issuecomment-508999581_

daboross:
This changes `LocalRoomPosition`'s serialized format from a kind-of-weird custom one to be that same kind-of-weird custom one when serializing to a binary format like bincode, but to be the regular format when serializing or deserializing from a "human readable" format like JSON or stdweb::Value.

In addition, this adds a call to `js_serializable!()` so `LocalRoomPosition`s can be passed directly into JavaScript code, and are serialized as objects.

Since serialization is now possible, this adds an example using `LocalRoomPosition` in the `MemoryReference::set` docs. This example should resolve #159.

One thing to note: the object produced by serializing `LocalRoomPosition` via `js!{}` won't be a `RoomPosition`, just an object with the same properties. This is documented in the `LocalRoomPosition` doc-comment.

### DIFF
--- a/screeps-game-api/src/positions.rs
+++ b/screeps-game-api/src/positions.rs
@@ -315,6 +315,45 @@ mod serde {
     js_serializable!(LocalRoomName);
 }
 
+/// This is a room position located in Rust memory.
+///
+/// It's "local" in the sense that while `RoomPosition` always references a room
+/// position allocated by and managed by the JavaScript VM, this is a
+/// self-contained plain-data struct in Rust memory.
+///
+/// # Using LocalRoomPosition
+///
+/// A `LocalRoomPosition` can be retrieved at any point by using
+/// [`RoomPosition::local`]. It can then be copied around freely, and have its
+/// values modified.
+///
+/// `&LocalRoomPosition` can be passed into any game method taking an object,
+/// and will be automatically uploaded to JavaScript as a `RoomPosition`.
+///
+/// If you need to manually create a `RoomPosition` from a `LocalRoomPosition`,
+/// use [`LocalRoomPosition::remote`].
+///
+/// # Serialization
+///
+/// `LocalRoomPosition` implements both `serde::Serialize` and
+/// `serde::Deserialize`.
+///
+/// When serializing, it will use the obvious format of `{roomName: String, x:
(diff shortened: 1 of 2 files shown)


screeps-game-api/src/positions.rs
```diff
@@ -396,16 +517,33 @@ mod room_pos_serde {
...
-                room_name: LocalRoomName {
-                    x_coord: data.room_x,
-                    y_coord: data.room_y,
-                },
-                x: data.x,
-                y: data.y,
-            })
+            if deserializer.is_human_readable() {
+                // an older version of the library always serialized as 'EfficientFormat' - this
+                // keeps compatibility with that.
```
ASalvail: Do you really want backward compatibility at the cost of lines of code to be maintained?

daboross: This is a good question!

I'm partial to backwards compatibility mostly because it's a pretty silent and easy to ignore change, and affects anyone storing LocalRoomPositions via `serde`?

We could definitely not do this, though, too.

ASalvail: I haven't started using it yet, so choose whatever you think is best.
I think that striving for backward compatibility in a pre-1.0 release is just asking for dead code, but I can see the reasoning.

daboross: Maybe we could make an issue for removing it next breaking release after this one?

I'm all in favor of breaking things, it's just nice to make them as visible as possible. I mean changing the serialization format is already technically a breaking change, it's just _slightly less breaking_ to accept the old one too on deserialization? ...

daboross:
Backwards compatibility code removed!

PR #186, opened 2019-07-21
Fix repository structure for splitting screeps-game-api and cargo-screeps

### LINKED ISSUE #181: Split repository into two repositories (under shared organization, maybe?)
Not sure if this needs to happen _now_, but it's something on the radar.

`cargo-screeps` and `screeps-game-api` are related, and work well together, but aren't fundamentally dependent on each other to function. I originally put them into the same repository because they were fairly small projects, and worked towards a common goal. But they've gotten bigger, and with more people using both they could probably use some more room. Splitting them, and possibly moving both under a new github organization, could be a good idea in the future.

There are other advantages to splitting them too, like not having to build both for pull requests which only really effect.

Biggest disadvantage I see is loosing the github issues which are shared on this repository. We could mitigate this by renaming `screeps-in-rust-via-wasm` to `screeps-game-api`, but then we'd still loose issues associated with `cargo-screeps`.

daboross:
This removes cargo-screeps from the repository, and merges the README and screeps-game-api's README into one file.

See also: https://github.com/rustyscreeps/cargo-screeps/pull/12

Closes #181.


README.md
```diff
@@ -18,12 +22,12 @@ Writing Screeps code in Rust can be nice, but it can also be annoying. If you ha
 other things you'd like to share, make an issue! We need to write more documentation, and if we have
 enough ideas, we can start an mdbook for this repository.
 
-- [cargo screeps usage docs](cargo-screeps/README.md)
-- [screeps-game-api api docs](https://docs.rs/screeps-game-api/)
+- [cargo screeps usage docs][cargo-screeps]
+- [screeps-game-api api docs][docs]
 - [screeps-starter-rust example project](https://github.com/daboross/screeps-starter-rust/)
```
ASalvail: A reminder to update after the move!

PR #189, opened 2019-07-22
Handle structure.my === undefined

daboross:
`my` appears to be used as a three-state boolean on owned structures! Expose this in the API, and fix errors when `my` is undefined.

Relevant engine code: https://github.com/screeps/engine/blob/a4e431c9ee33dcc55c04f3adc479145c210c73fd/src/game/structures.js#L219


snorrwe:
I did encounter panics because of this, so this is definitely welcome

daboross:
> Good catch. Did you have an issue with that? I wonder which `OwnedStructure` could have no owners... I mean, this is meant to be an invariant for this trait...

Yeah - ran into it testing scouting. I think the best (and maybe only?) example is `StructureController`, which is "owned" but doesn't always have an owner. Might also apply to the extractors in center rooms, but I'm not sure.

PR #190, opened 2019-07-22
Make StructurePortal not OwnedStructure

### LINKED ISSUE #120: StructurePortal probably shouldn't be Attackable?
See:

https://github.com/daboross/screeps-in-rust-via-wasm/blob/a6d498dd558f83063f6b3506bf237cee8c38bc17/screeps-game-api/src/objects/mod.rs#L399

Even if they have hits, it doesn't make sense to attack them. And if they're like `StructureController`, they might not have hits at all anyways. I haven't tested whether they actually have hits in game.

I'll get to this eventually but I just wanted to note the issue while working on other stuff.

daboross:
Fixes #120.


src/objects/structure.rs
```diff
@@ -130,7 +130,6 @@ impl Structure {
             Structure::Observer(v) => Some(v),
             Structure::PowerBank(v) => Some(v),
             Structure::PowerSpawn(v) => Some(v),
-            Structure::Portal(v) => Some(v),
             Structure::Rampart(v) => Some(v),
```
ASalvail: That looks like a non-comprehensive pattern matching.

src/objects/structure.rs
```diff
@@ -164,7 +163,7 @@ impl Structure {
             Structure::Observer(v) => Some(v),
             Structure::PowerBank(v) => Some(v),
             Structure::PowerSpawn(v) => Some(v),
-            Structure::Portal(v) => Some(v),
+            Structure::Portal(v) => None,
```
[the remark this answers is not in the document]
daboross: :+1:

src/objects/structure.rs
daboross: I thought I'd actually compiled this code...

PR #193, opened 2019-07-22
Build on stable rust, and remove multiple OS CI

daboross:
Since we're always cross-compiling for wasm32-unknown-unknown, there is very little we can do to mess up on Windows or OSX but not Linux.

Separately, screeps-game-api currently supports stable rust, so testing on that seems like a good idea!

This removes OSX and Windows continuous integration, and adds testing for stable & beta rust on linux.


.travis.yml
```diff
@@ -1,10 +1,11 @@
 language: rust
 sudo: false
 cache: cargo
-rust: nightly
-os:
-- linux
-- osx
+rust:
+- stable
+- beta
```
ASalvail: Up to you, but I suspect that we'd catch it on nightly if something was going to break?

daboross: I _think_ it's helpful, mainly in case a breaking change makes it into beta but is fixed in nightly? Like if we always succeed betas, we will always succeed on the next stable release - and that guarantee isn't necessarily there for nightly rust.

Not super needed, but it could be nice to have.

PR #194, opened 2019-07-23
Add a way to retrieve both keys and values of game objects

ASalvail:
Adds a way to access the whole mapping of `String -> type` of the game objects. This prevents having to call the method twice to get both the keys and the values, when needed.

Also splits `Game` into submodules properly.


src/macros.rs
```diff
@@ -451,9 +451,15 @@ macro_rules! game_map_access {
                     #[$attr]
             )*
             pub mod $mod_name {
+                use std::collections::HashMap;
+
                 use crate::objects;
                 use crate::macros::*;
 
+                /// Retrieve the full `HashMap<String, $type>`.
```
daboross: Want to use a `#[doc = stringify!($type)]` hack to make this work?

We'd probably need to add a helper like this one:
https://github.com/rustyscreeps/screeps-game-api/blob/18fffc58b0c048cba680e10394530b3dfc2fb728/src/macros.rs#L348-L364

but then we could do something like `#[doc = concat!("Retrieve the full `HashMap<String, ", stringify!($type), ">`.")]`.

ASalvail: Thanks for the solution, I hadn't expected this not to work!

PR #196, opened 2019-07-23
Misc. doc improvements

daboross:
I think this should fix most if not all broken links, and has some other small improvements as well.

The docs still aren't entirely consistent nor consistently useful, but this should improve that a small bit.

### DIFF
--- a/src/pathfinder.rs
+++ b/src/pathfinder.rs
@@ -1,3 +1,14 @@
+//! Manually accessing the [`PathFinder`] API
+//!
+//! This contains functionality from the [`PathFinder`] object in Screeps, which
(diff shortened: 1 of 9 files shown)


src/pathfinder.rs
```diff
@@ -1,3 +1,14 @@
+//! Manually accessing the [`PathFinder`] API
+//!
+//! This contains functionality from the [`PathFinder`] object in Screeps, which
+//! is itself a binding to a C++ implementation of the astar algorithm backed by
```
ASalvail: Isn't it a jump-point algorithm?

daboross: Ah - I hadn't heard of that algorithm - it totally is!

PR #197, opened 2019-07-24
Move 'pub mod x;' outside of game_map_access! macro

daboross:
This moves the documentation and `pub mod x;` outside of  `game_map_access!`, so it only generates function names.

My main reason for doing this would be to work around https://github.com/intellij-rust/intellij-rust/issues/3710, where IntelliJ-Rust doesn't / can't figure out macros which create `pub mod` statements. I don't personally use IntelliJ, but it is an annoyance for some people who do. It doesn't seem like they'll be fixing that bug anytime soon, and it's a fairly simple workaround.

I don't think this should be a large change - and it shouldn't be _too_ more more code to maintain.

### DIFF
--- a/src/game.rs
+++ b/src/game.rs
@@ -17,34 +17,49 @@ pub mod map;
 pub mod market;
 pub mod shards;
 
-game_map_access! {
-    /// See [http://docs.screeps.com/api/#Game.constructionSites]
-    ///
-    /// [http://docs.screeps.com/api/#Game.constructionSites]: http://docs.screeps.com/api/#Game.constructionSites
-    (construction_sites, objects::ConstructionSite, Game.constructionSites),
-    /// See [http://docs.screeps.com/api/#Game.creeps]
-    ///
-    /// [http://docs.screeps.com/api/#Game.creeps]: http://docs.screeps.com/api/#Game.creeps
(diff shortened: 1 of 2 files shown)


src/game.rs
```diff
@@ -17,34 +17,49 @@ pub mod map;
...
+}
+
+/// See [http://docs.screeps.com/api/#Game.flags]
+///
+/// [http://docs.screeps.com/api/#Game.flags]: http://docs.screeps.com/api/#Game.flags
+pub mod flags {
+    game_map_access!(objects::Flag, Game.flags);
+}
+
+// TODO: See [http://docs.screeps.com/api/#Game.resources]
```
daboross: Sidenote: I have some code to fix this, but since it's not really tied to this change beyond affecting the same region of code, I'm planning on submitting it in another PR.

src/game.rs
```diff
@@ -17,34 +17,49 @@ pub mod map;
...
+///
```
ASalvail: Not needed

daboross: I didn't change it in this PR, but it should be a double-`//` comment. The line below this one mentions `Game.resources`, and I assume it was intended as part of the TODO.

Issue #198, opened 2019-07-25
Integrate `LocalRoomName` into APIs better


daboross:
It'd be nice if we could get rooms via `LocalRoomName` directly, or at least do something nicer than `game::rooms::get(&name.to_string())`.

May or may not want to wait on #191.

If implemented by creating the string in Rust code, using a  stack-based string could be more efficient than `to_string()`. An appropriately sized [`arrayvec::ArrayString`](https://docs.rs/arrayvec/0.4.11/arrayvec/struct.ArrayString.html) should work. (we should be able to just `write!(string, "{}", name);` if it's of an OK capacity).

Issue #191, opened 2019-07-22
Use 32bit representation for room names and room positions


daboross:
Ever since https://github.com/screeps/engine/pull/87, screeps has internally represented RoomPositions as 32 bit integers.

It isn't strictly guaranteed that the world is small enough to do this, but it's as close to a guarantee as we'll get. I think it should be reasonable to represent `LocalRoomPosition`s locally as the same 32 bit integer, and we should be able to switch `LocalRoomName` to use that representation as well (maybe exactly the same, with 0,0 room coord so we can easily add room name to room coordinates).

If we do this, then it'd probably be reasonable to use `LocalRoomName` and `LocalRoomPosition` in more places. Converting `RoomPosition` to `LocalRoomPosition` would become a single integer copy, so we could probably use `LocalRoomPosition` for everything except sending positions into methods that require the `RoomPosition` prototype, like `Creep::move_to`.

ASalvail:
Honestly, with a world that has 256x256 rooms, the tick speed would be so slow that I think it's safe to assume it'll never be reached. However, wouldn't it be simpler to put that integer in `RoomPosition` itself? Or maybe implement `Deref<RoomPosition> for LocalRoomPosition`?

daboross:
I think it can be useful to have the two representations, even if just for the sake of having `#[derive(ReferenceType)]` make the right conversion code for `RoomPosition`. We can't use `#[derive(ReferenceType)]` on something which doesn't actually store a reference to JS, and if it stores a reference to JS it will be more expensive to move around.

I mean, here are some differences we could only keep if we have separate types:

- `RoomPosition` can be `AsRef<Reference>` due to storing a reference internally
- when converting to JS, `RoomPosition` is already a structure with a prototype
- `RoomPosition` gets very nice implementations of all the conversion types because of `#[derive(ReferenceType)]`
- `RoomPosition` deserializes from JavaScript if and only if the JS side is a `RoomPosition` (side effect of `#[derive(ReferenceType)]`)
- `RoomPosition` can allow using the 'pos' property from an object as an argument to another method without ever needing to create a `new RoomPosition`.
- `LocalRoomPosition` can be `Copy` because it doesn't store a reference
- `LocalRoomPosition` can have cheap addition because it has a local number
- `LocalRoomPosition` can deserialize from any object with `x`,`y` and `roomName` properties

These different use cases are possible because the internal representations are different.

[...]

PR #200, opened 2019-07-26
Flip N/S signs in LocalRoomPosition

daboross:
This is a preliminary to work in #198 and #191 that I want to get out of the way.

Previously, "north" was considered positive, and "south" was considered negative. This is opposite of the direction for room positions, and also opposite what the game's packed representation uses for room positions.

Reason I'm pulling it out from the other changes is that it's a hard and silent breaking change to anyone using the coordinates stuff on `LocalRoomName`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased
 ==========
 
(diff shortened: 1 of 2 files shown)


daboross:
Changelog entry added! Thanks for the reminder!

PR #201, opened 2019-07-26
Add forgotten changelog entries

ASalvail:
I thought we had been a bit more thorough in keeping track of changes in the changelog, but I suppose we forgot to keep track of those for the main game API. Here is a PR to fix that.


ASalvail:
Yours is more complete, closing this PR.

PR #203, opened 2019-07-28
Improve LocalRoomName and LocalRoomPosition (1/2)

daboross:
This is the first half of my changes to `LocalRoomName` and `LocalRoomPosition`, aimed at implementing #191.

This PR:
- Splits `LocalRoomName` and `LocalRoomPosition` into separate files
- Includes some small-but-necessary improvements to the  `LocalRoomName` API
  - we can now output to an ArrayString
  - the parsing API is simplified and the `str::parse()` method can be used
- Changes `LocalRoomPosition`'s internal representation to be a single u32 (as discussed in #191)
  - switching between `LocalRoomPosition` and `RoomPosition` now translates a single `i32`
- Adds a set of pure-rust math utilities to `LocalRoomPosition`

I'm sending this as its own PR because unifying `LocalRoomPosition` and `RoomPosition` seems like a hard change, and I believe these changes will be useful regardless.

For reviewing, it might be helpful to look at:
- changes from before the file split: https://github.com/rustyscreeps/screeps-game-api/pull/203/files/6e2278d647d5db04d3bd0dfd3ad17bfb74784ba0
- the file split: https://github.com/rustyscreeps/screeps-game-api/pull/203/commits/de6a79156cdbeb63c0ce43358d42e6a374aae5ac
- changes after the file split: https://github.com/rustyscreeps/screeps-game-api/pull/203/files/de6a79156cdbeb63c0ce43358d42e6a374aae5ac..e3cb588b8f9cccaa26df3b38dc8acf7b1d5d6c0c

I could split this into more PRs (improve local room name, then split, then improve local room position), but it seemed like it'd just take more time than use gained. If you think it'd be a good idea, I can do that.


src/positions.rs
```diff
@@ -224,39 +224,58 @@ impl<'a> fmt::Display for LocalRoomNameParseError<'a> {
...
+impl PartialEq<str> for LocalRoomName {
+    fn eq(&self, other: &str) -> bool {
+        let s = self.to_array_string();
+        s.eq_ignore_ascii_case(other)
+    }
+}
+impl PartialEq<LocalRoomName> for str {
+    #[inline]
+    fn eq(&self, other: &LocalRoomName) -> bool {
+        <LocalRoomName as PartialEq<str>>::eq(other, self)
```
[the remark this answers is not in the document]
daboross: I'll impl `Eq`!

The reason I'm doing the explicit reference here is that we have all these other implementations for `&str`, `&&str`, etc., and I want to make sure that I don't end up creating an infinite loop. This one might be alright without it, or with some other annotations, but then the code could change in the future and it'd be super easy with all of these implementations to end up creating an infinitely recursive function. Will create a comment to give these motivations!

src/local/room_name.rs
```diff
@@ -0,0 +1,346 @@
...
+
+/// A structure representing a room name.
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
+pub struct LocalRoomName {
+    /// Inner x coordinate representation.
+    ///
+    /// 0 represents E0, positive numbers represent E(x)
+    ///
+    /// -1 represents W0, negative numbers represent W((-x) - 1)
+    pub x_coord: i32,
```
ASalvail: Maybe consider leaving those private? This is exposing a lot of the inner structure of something that is meant to be a _name_.

src/local/room_name.rs
```diff
@@ -0,0 +1,346 @@
...
+
+impl LocalRoomName {
+    /// Creates a new room name from the given input.
+    ///
+    /// This will parse the input, and return an error if it is in an invalid
+    /// format.
+    #[inline]
+    pub fn new<T>(x: &T) -> Result<Self, LocalRoomNameParseError>
+    where
+        T: AsRef<str> + ?Sized,
```
ASalvail: What is this expecting? A `&str`? A `T: HasPosition`? Doc is unclear (and I honestly can't figure it out by just looking at the signature).

src/positions.rs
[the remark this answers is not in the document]
daboross: Oh! I forgot - `Eq` can only be applied if the left and right sides are the same type. `LocalRoomName` does implement `Eq<LocalRoomName>`, but it can't implement it for `str`.

src/local/room_name.rs
```diff
@@ -0,0 +1,346 @@
...
+    {
+        x.as_ref().parse()
+    }
+
+    /// Creates a new room name from the given position parameters.
+    #[inline]
+    pub fn from_coords(east: bool, south: bool, x_pos: i32, y_pos: i32) -> Self {
```
ASalvail: Consider making this `pub(crate)`

src/local/room_position.rs
```diff
@@ -19,104 +24,258 @@ use crate::objects::{HasPosition, RoomPosition};
...
 /// macro provided by `stdweb`, or helper methods using the same code like
-/// [`MemoryReference::set`][crate::memory::MemoryReference::set].  It will be
+/// [`MemoryReference::set`][crate::memory::MemoryReference::set]. It will be
 /// serialized the same as in JSON, as an object with `roomName`, `x` and `y`
 /// properties.
 ///
 /// *Note:* serializing using `js!{}` or `MemoryReference::set` will _not_
-/// create a `RoomPosition`, only something with the same properties. Use
-/// `.remote()` if you need a `RoomPosition`.
+/// create a `LocalRoomPosition`, only something with the same properties.
```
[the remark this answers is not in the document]
daboross: Yep, this is incorrect...

src/local/room_position.rs
```diff
@@ -19,104 +24,258 @@ use crate::objects::{HasPosition, RoomPosition};
...
+///        let pos = pos_from_packed(@{pos.packed_repr()})
+///    };
+///    # }
+///    ```
 ///
 /// [`bincode`]: https://github.com/servo/bincode
-#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
+/// [`RoomPosition::local`]: crate::RoomPosition::local
+#[derive(Copy, Clone, Eq, PartialEq, Hash)]
+#[repr(transparent)]
```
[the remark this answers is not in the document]
daboross: The `#[repr(transparent)]` attribute guarantees it's represented in memory equivalently to its only field. Just guarantees something that should already be true? Doesn't affect serde or stdweb representation.

src/local/room_name.rs
```diff
@@ -0,0 +1,346 @@
...
+        let mut res = ArrayString::new();
+        write!(res, "{}", self).expect("expected ArrayString write to be unfallible");
+        res
+    }
+}
+
+impl ops::Add<(i32, i32)> for LocalRoomName {
+    type Output = Self;
+
+    /// Adds an (x, y) coordinate pair to this room name.
```
ASalvail: This again assumes that users are aware of the implementation.

src/local/room_position/pair_utils.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /// Returns this positions in-room coordinates as a pair of signed integers.
+    #[inline]
+    pub fn coords_signed(&self) -> (i32, i32) {
+        (self.x() as i32, self.y() as i32)
+    }
+}
+
+impl Into<(u8, u8)> for LocalRoomPosition {
```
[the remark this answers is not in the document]
daboross: I tried to, but apparently `(u8, u8)` is one to many layers of wrapping to implement a trait for.

I'll make a comment about this, though.

src/local/room_position.rs
```diff
@@ -19,104 +24,258 @@ use crate::objects::{HasPosition, RoomPosition};
...
+    ///
+    /// For `Wxx` rooms, `room_x = -xx - 1`. For `Exx` rooms, `room_x = xx`.
+    ///
+    /// For `Nyy` rooms, `room_y = -yy - 1`. For `Syy` rooms, `room_y = yy`.
+    ///
+    /// This is the same representation used in the Screeps server, allowing for
+    /// easy translation. Besides the method names and display representation,
+    /// this is the one part of LocalRoomPosition copied directly from the
+    /// engine code.
+    packed: u32,
```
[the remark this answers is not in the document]
daboross: I think `i32` would cause a whole other set of conversions when getting coordinates?

The main reason two reasons I chose `u32` were:
-  the bitwise operators act nicely for bitpacking. `i32`'s  `>>` operator will sign-extend, which could be subtly wrong?
- none of the values stored in this bitwise repr are ever negative - even the world position

Somewhat differently, I did u32 over `[u8; 4]` mainly for ease of translating to JS. `packed as i32` should be free, and handle the values correctly.

src/local/room_position.rs
```diff
@@ -0,0 +1,375 @@
...
+use std::fmt;
+
+use super::LocalRoomName;
+
+mod extra_math;
+mod game_math;
+mod pair_utils;
+mod world_utils;
+
+const HALF_WORLD_SIZE: i32 = 128;
```
ASalvail: What is this meant to represent? I thought it was to wrap around from `i8` to `u8`.

src/local/room_position.rs
```diff
@@ -0,0 +1,375 @@
...
+    }
+}
+
+impl LocalRoomPosition {
+    /// Create a new LocalRoomPosition
+    ///
+    /// # Panics
+    ///
+    /// Will panic if either `x` or `y` is larger than 49, or if `room_name` is
+    /// outside of the range `E127N127 - W127S127`.
```
ASalvail: I believe the first pair is 128?

src/local/room_position.rs
```diff
@@ -19,104 +24,258 @@ use crate::objects::{HasPosition, RoomPosition};
...
+    pub fn new(x: u32, y: u32, room_name: LocalRoomName) -> Self {
+        assert!(x < 50, "out of bounds x: {}", x);
+        assert!(y < 50, "out of bounds y: {}", y);
+        assert!(
+            (-128..=127).contains(&room_name.x_coord),
+            "out of bounds room_x: {}",
+            room_name.x_coord,
+        );
+        assert!(
+            (-128..=127).contains(&room_name.y_coord),
```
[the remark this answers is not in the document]
daboross: I was thinking of them as different concepts (packed repr magic number vs. accepted range), but I guess that makes sense. Will change that.

src/local/room_position.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    #[inline]
+    pub fn from_packed(packed: i32) -> Self {
+        LocalRoomPosition {
+            packed: packed as u32,
+        }
+    }
+
+    #[inline]
+    fn room_x(self) -> i32 {
```
ASalvail: Without looking at the code, it's unclear whether `room_x` is the x coordinates of the room, or the x coordinate within a room.

src/local/room_position.rs
```diff
@@ -0,0 +1,375 @@
...
+    use crate::{
+        macros::*,
+        objects::{HasPosition, RoomPosition},
+        traits::{TryFrom, TryInto},
+    };
+
+    use super::LocalRoomPosition;
+
+    impl LocalRoomPosition {
+        pub fn remote(self) -> RoomPosition {
```
ASalvail: Given it is implemented here in a private module, I doubt it'll get included in the docs.

src/local/room_position.rs
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    impl crate::traits::FromExpectedType<Reference> for LocalRoomPosition {
+        fn from_expected_type(reference: Reference) -> Result<Self, crate::ConversionError> {
+            Self::try_from(Value::Reference(reference))
+        }
+    }
+
+    impl HasPosition for LocalRoomPosition {
+        fn pos(&self) -> RoomPosition {
```
ASalvail: (For later) maybe consider having `pos` return a `LocalRoomPosition`. Then, it may make more sense to have `RoomPosition` be `RemoteRoomPosition`.

src/local/room_position/game_math.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub fn in_range_to(self, target: LocalRoomPosition, range: u32) -> bool {
+        self.get_range_to(target) < range
+    }
+
+    #[inline]
+    pub fn is_equal_to(self, target: LocalRoomPosition) -> bool {
+        self == target
+    }
+
+    /// True if this position is in the same room as and is within 1
```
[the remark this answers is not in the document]
daboross: I mean I like my slightly-ambiguous English, but that is more clear.

src/positions.rs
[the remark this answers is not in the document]
ASalvail: Ah, I didn't know that. I suppose it makes sense 🤔

src/local/room_name.rs
daboross: Good idea!

I didn't make any changes to the representation of `LocalRoomName` including visibility since I thought that might happen at a later date, but no reason why we can't make it private now.

src/local/room_name.rs
daboross: Fair! I'll update the docs. It's expecting something that's string-like - `&str`, `&String`, `&ArrayString<[u8; N]>`, `Cow<'_, str>`, etc.

src/local/room_name.rs
daboross: Will do.

src/local/room_position.rs
daboross: It's a magic number copied from the JavaScript implementation, mostly. I've added a comment.

src/local/room_position.rs
daboross: The maximum room coordinate is 128, but the fact of there being multiple "0" rooms (`E0N0 != W0N0`) means the actual room name of both maximum and minimum is `127`.

src/local/room_position.rs
```diff
@@ -12,8 +12,19 @@ mod game_math;
...
+/// First, this is the constant added to room coordinates before they're stored
+/// in the packed representation.
+///
+/// Second, `-HALF_WORLD_SIZE` is the minimum representable room name
+/// coordinate, and `HALF_WORLD_SIZE - 1` is the maximum representable room name
+/// coordinate.
 const HALF_WORLD_SIZE: i32 = 128;
 
+/// Valid room name coordinates.
+const VALID_ROOM_NAME_COORDINATES: Range<i32> = (-HALF_WORLD_SIZE..HALF_WORLD_SIZE);
```
ASalvail: Oh, that's neat!

src/local/room_name.rs
ASalvail: Up to you :)

src/local/room_name.rs
ASalvail: Shouldn't it `Deref<&str>` or something?

src/local/room_position.rs
daboross: Documentation added!

src/local/room_position.rs
daboross: Planning on having this change in the next round, once we get rid of the "remote" room position.

src/local/room_name.rs
```diff
@@ -84,7 +87,11 @@ impl LocalRoomName {
 impl ops::Add<(i32, i32)> for LocalRoomName {
     type Output = Self;
 
-    /// Adds an (x, y) coordinate pair to this room name.
+    /// Offsets this room name by a given horizontal and vertical (x, y) pair.
+    ///
+    /// The first number offsets to the west when negative and to the east when
+    /// positive. The first number offsets to the north when negative and to
```
ASalvail: second

ASalvail: Would it be more natural to inverse the y direction to follow the cartesian standard?

daboross: #200 :p

I would argue the game set this standard when they made `north` negative in in-room coordinates?

src/local/room_position.rs
daboross: I guess it isn't too intuitive, but `impl Type` blocks are always applicable, no matter the location. Many of the utility methods are implemented in private modules as well?

src/local/room_name.rs
daboross: It could be, but I think `AsRef<&str>` is strictly more general?

For example, `Deref<&str>` is implemented for `String`, but not for `&String`.

src/local/room_name.rs
daboross: Documentation added.

src/local/room_position.rs
ASalvail: So they inherit the struct's visibility? Interesting (and easier for when you want to split the implementation like in this case).

src/local/room_name.rs
ASalvail: Yeah, that's from game design. Pixels are counted from the top left 😞

src/local/room_position.rs
daboross: As I understand it, `impl` blocks are always applicable, no matter where they're located in submodules. It's even technically possible to write `impl` blocks within functions, and in there they behave exactly like anywhere else.

In this PR I've implemented most of the methods in similar private modules, just ones declared at the top of the file and written in separate files.

src/local/room_position.rs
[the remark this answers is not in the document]
daboross: Fixed

daboross:
Alright - most small comments addressed.

I'm also slightly concerned about the world-coords vs. in-room coords difference, but I'm not sure about the best way to resolve it.

I think having the `Add` and `Sub` operations work on world coordinates makes sense, because it's more general than the room coordinates - it's still possible to modify in-room coords using those implementations.

I'm less sure about the `Into` implementations. Since they're present on `RoomPosition`, I was just going to keep them on `LocalRoomPosition` too.

src/local/room_name.rs
ASalvail: I'm blind: I noticed the `AsRef`, but not the `str`. It makes so much more sense now!

ASalvail:
> I'm also slightly concerned about the world-coords vs. in-room coords difference, but I'm not sure about the best way to resolve it.
> 
> I think having the `Add` and `Sub` operations work on world coordinates makes sense, because it's more general than the room coordinates - it's still possible to modify in-room coords using those implementations.
Well, you can add and sub on room coordinates via the `RoomName`.

Unless we're talking about within room coordinates, then I agree, using world coordinates is just the more general case.

> I'm less sure about the `Into` implementations. Since they're present on `RoomPosition`, I was just going to keep them on `LocalRoomPosition` too.

Well, they're just shortcuts to getting both `x` and `y` at once. Whether they should be `Into` or just use a method is mostly a matter of semantics. The one thing I dislike is that they don't follow the invariant of `impl Into<T> for T: From<T>`.

Otherwise, all looking good!

daboross:
Alright! I was talking about world coordinates.

I'll merge this!

PR #206, opened 2019-07-28
Merge RoomPosition and LocalRoomPosition into Position (2/2)

daboross:
This is dependent on #203, and thus all but the last 6 commits are copied from that PR. I haven't merged that PR yet, as there's still some discussion and open comment chains, but I'm opening this one in parallel since it's finished too.

This changes all game methods concerned with `RoomPosition`s to use `LocalRoomPosition` instead, then removes `RoomPosition` and renames `LocalRoomPosition` to `RoomPosition`.

I've written each of the 5 commits in this PR so that they're fairly independent. Each passes tests without the others, and represents a fairly atomic change.


src/pathfinder.rs
```diff
@@ -349,17 +349,17 @@ where
...
     let goals_js: Reference = js_unwrap!(@{goals});
-    search_real(&origin.pos(), &goals_js, opts)
+    search_real(origin.pos(), &goals_js, opts)
 }
 
 scoped_thread_local!(static PF_CALLBACK: &'static dyn Fn(String) -> Reference);
 
 fn search_real<'a, F>(
-    origin: &RoomPosition,
+    origin: LocalRoomPosition,
```
ASalvail: We should consider making `LocalRoomPosition` `Copy` (if it isn't already).

src/objects/impls/room_position.rs
```diff
@@ -156,57 +156,57 @@ impl RoomPosition {
...
+impl Into<(u8, u8)> for RemoteRoomPosition {
     fn into(self) -> (u8, u8) {
         (self.x() as u8, self.y() as u8)
     }
 }
 
-impl Into<(u16, u16)> for RoomPosition {
+impl Into<(u16, u16)> for RemoteRoomPosition {
     fn into(self) -> (u16, u16) {
         (self.x() as u16, self.y() as u16)
```
ASalvail: It might be best to make it local once and get `x` and `y` from it. Saves a js-rust transport.

src/pathfinder.rs
ASalvail: Just saw that it was!

src/local/room_position/game_methods.rs
```diff
@@ -0,0 +1,97 @@
...
+
+impl LocalRoomPosition {
+    pub fn create_construction_site(self, ty: StructureType) -> ReturnCode {
+        js_unwrap!(
+            pos_from_packed(@{self.packed_repr()})
+                .createConstructionSite(__structure_type_num_to_str(@{ty as u32}))
+        )
+    }
+
+    pub fn create_named_construction_site(self, ty: StructureType, name: &str) -> ReturnCode {
```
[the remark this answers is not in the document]
daboross: Main reason is that `LocalRoomPosition` is the size of `u32`, and thus a reference to it will be the exact same size, and just require more indirection to use.

Since `LocalRoomPosition` is `Copy`, it should still work for `&LocalRoomPosition` - only difference is that no reference is needed.

I first thought about this when I encountered the [trivial_copy_pass_by_ref clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref).

src/objects/impls/room_position.rs
daboross: Eh- I can change this!

I did the transition commit first mostly to ensure that we can cleanly revert removing `RemoteRoomPosition` if it's deemed necessary in the future, without having to first disambiguate it and `LocalRoomPosition`.

src/local/room_position/game_methods.rs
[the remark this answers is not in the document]
ASalvail: Oh that makes sense. I wasn't sure whether `Self` would also be taken by copy. Would it be worth changing the others also?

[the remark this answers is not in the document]
daboross: Definitely! I, uh, thought I had. Must have missed that!

daboross:
Rebased on master, and fixed some of the concerns!

daboross:
`RoomPosition` renamed! Had to rebase to update the CHANGELOG entries correctly.

src/local/room_position.rs
```diff
@@ -28,18 +28,26 @@ const VALID_ROOM_NAME_COORDINATES: Range<i32> = (-HALF_WORLD_SIZE..HALF_WORLD_SI
...
 /// Represents a position in a particular room in Screeps.
 ///
-/// This is "local" in the sense that while other structures alway references
-/// things allocated by and managed by the JavaScript VM, this is a
-/// self-contained plain-data struct in Rust memory, the same size as a `i32`.
+/// **Note:** This is analagous to the `RoomPosition` JavaScript type.
+///
+/// We've renamed this type to `Position` in `screeps-game-api` to reflect the
+/// fact that it's implemented entirely as a local type, and does represent a
+/// position within the global world.
```
ASalvail: Well, limited to a shard.

daboross: Fixed!

PR #209, opened 2019-08-03
Rework LocalRoomName to use a packed repr similar to Position

daboross:
This should make more efficient to store `LocalRoomName`s, to retrieve `LocalRoomName`s from `Position`s, and to create `Position` from `LocalRoomName`s.

### DIFF
--- a/src/local/room_name.rs
+++ b/src/local/room_name.rs
@@ -7,21 +7,23 @@ use std::{
 
 use arrayvec::ArrayString;
 
+use super::{HALF_WORLD_SIZE, VALID_ROOM_NAME_COORDINATES};
+
 /// A structure representing a room name.
 #[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
 pub struct LocalRoomName {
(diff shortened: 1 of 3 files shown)


src/local/room_name.rs
```diff
@@ -64,13 +70,54 @@ impl LocalRoomName {
...
     #[inline]
-    pub(crate) fn from_coords(east: bool, south: bool, x_pos: i32, y_pos: i32) -> Self {
-        LocalRoomName {
-            x_coord: if east { x_pos } else { -x_pos - 1 },
-            y_coord: if south { y_pos } else { -y_pos - 1 },
+    pub(crate) fn from_packed(packed: u16) -> Self {
+        LocalRoomName { packed }
+    }
+
+    /// Creates a new room name from room coords with direction implicit in
```
ASalvail: Just learned that `cargo doc --include-private` is a thing that can be very helpful for us!

daboross: It's nice!

PR #211, opened 2019-08-11
Add #[inline] to many constant functions

### LINKED ISSUE #64: Consider marking constant functions `#[inline]`
Things like `Part::cost()` are constant functions on constant values which could probably be completely eliminated from constant expressions if allowed to inline. LTO should allow them too anyways, but it'd be nice to give the compiler a bit more freedom.

This is pure optimization, but I was thinking about some code I'm writing, and this seemed like a good idea to do in the future. The main hard part is going through `constants.rs` and deciding what functions are good to inline and what aren't.

daboross:
This should just open up optimization possibilities. The biggest downside might be increased compilation times since now it'll be possible to inline these when compiling downstream projects, but I don't believe that will have too large of an impact relative to current compile times.

I've only marked functions which do constant, cheap calculations as `#[inline]` here, since I think those are the ones which could most easily be completely optimized away.

Closes #64.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Unreleased
 - Implement `Clone` for `Structure`
 - Split [cargo-screeps](https://github.com/rustyscreeps/cargo-screeps/) out into a separate
   repository
+- Mark most constant functions as inlinable (#221)
 - Misc. documentation improvements.
 
 [pos-doc]: https://docs.rs/screeps-game-api/0.6.0/screeps/local/struct.Position.html
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -102,6 +102,7 @@ impl ReturnCode {
     ///
     /// `ReturnCode::Ok` is turned into `Result::Ok`, all other codes are turned
     /// into `Result::Err(code)`
+    #[inline]
     pub fn as_result(self) -> Result<(), Self> {
         match self {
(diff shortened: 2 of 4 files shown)


daboross:
It does feel  a bit bad, but I feel better knowing that this just opens up the possibility of inliing, and doesn't force the compiler to do it. I think all this annotation should do is make rustc include the AST for these functions in the output library- so they can be more easily inclined as rust functions rather than as part of linking with lto?

daboross:
I'm not exactly sure why it won't automatically choose to do this for small functions, but maybe it's just to do with how much more bloated a library could become and having most inlining metrics analyzing both the called function and the caller?

ASalvail:
From what I read, without the user willingly buying into potential cross-crate optimization, it just won't happen without `#[inline]`.

Adding it is harmless (except maybe for compilation time) and potentially helpful if the user isn't using _lto_. I wouldn't go out of my way to add `#[inline]`, but I guess they're fine for this PR. Refusing them would mean to try and remove the other ones too, which I'm not inclined to do.

daboross:
That sounds right to me - and I guess it is reasonably harmless to merge this, assuming we don't go and delete all the others.

If we want to in the future, it shouldn't be too hard to just delete all lines which are `<whitespace>#[inline]\n` from the repo.

PR #215, opened 2019-08-11
Make Bodypart fields public

### LINKED ISSUE #214: Fields on Bodypart should be marked as pub
Currently Bodypart is defined as:

```rust
#[derive(Clone, Debug)]
pub struct Bodypart {
    boost: Option<ResourceType>,
    part: Part,
    hits: u32,
}
```

Which prevents access to any of the fields. These should be marked as public so this struct can be used.

daboross:
Fixes #214.

I'm not 100% sure if there was a reason we had them private to begin with? If there was, then maybe this could be addressed with accessor methods instead.

I've added a `_non_exhaustive` field to make it a non-breaking change to add new properties (not that that matters _that much_).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Unreleased
 - `StructurePortal` no longer implements `OwnedStructure` and `Attackable`. (#190)
 - Collections provided by `Game` now implement the `hashmap` function to retrieve both keys
   and values at once. (#194)
+- Allow accessing fields of the `Bodypart` struct. (#215)
 - Implement `Clone` for `Structure`
 - Split [cargo-screeps](https://github.com/rustyscreeps/cargo-screeps/) out into a separate
   repository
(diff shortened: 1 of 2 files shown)


ASalvail:
I don't get the `_non_exhaustive()` hack?
Regardless, adding new public variables can't break anything, can it?

daboross:
Yeah, I was referring to adding new public fields of `Bodypart` in the future, like if they added a mechanic like boosting parts, but with with power. Or any other change to that struct?

The only reason the `_non_exhaustive: ()` field is there is that if a struct has only public fields, it would be a breaking change to add a new one since then old code doing `let Bodypart { boost, part, hits } = bodypart;` would no longer capture all fields and would error. Another case would be constructing it - this prevents `let x = Bodypart { boost: x, part: y, hits: z }`, which would also succeed but fail to compile if we ever add fields.

---

I don't have any fields in particular I want to add in the future, but this would let us do that if we ever want to. I would only consider doing it if the JavaScript representation added fields first.

PR #218, opened 2019-08-11
Last `LocalRoomName` changes: use in all APIs, rename to RoomName

### LINKED ISSUE #198: Integrate `LocalRoomName` into APIs better

daboross:
This makes all public functions which refer to room names use our in-rust `RoomName` representation. They still translate the room names to strings before transferring to the API (`RoomName` implements `JsSerialize` using its `Serialize` impl, which allocates a stack-based string and serializes that), but that could change in the future?

This also renames `LocalRoomName` to `RoomName`, since it's no longer an optional "local" version of something but the canonical structure to use for room names. I do still like the name `RoomName` most, because even if it isn't the most technically accurate, it matches the wording the screeps documentation uses, and fields like `RoomPosition.roomName` in JS. Even if it represents a room coordinate, there's value in being consistent with JS naming conventions. I'm thinking of the resources in slack and people who know to use "room name" to refer to this bit of info.

Resolves #198.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -125,7 +125,7 @@ js_deserializable!(ReturnCode);
 /// ```no_run
 /// use screeps::{find, game, Room};
 ///
-/// let room: Room = game::rooms::get("E23S55").unwrap();
+/// let room: Room = game::rooms::get("E23S55".parse().unwrap()).unwrap();
 ///
 /// let creeps = room.find(find::CREEPS);
 /// # let _ = creeps;
--- a/src/game.rs
+++ b/src/game.rs
@@ -45,7 +45,29 @@ pub mod flags {
 ///
 /// [http://docs.screeps.com/api/#Game.rooms]: http://docs.screeps.com/api/#Game.rooms
 pub mod rooms {
-    game_map_access!(objects::Room, Game.rooms);
+    use std::collections::HashMap;
+
+    use crate::{local::LocalRoomName, macros::*, objects::Room};
+
+    /// Retrieve the full `HashMap<LocalRoomName, Room>`.
+    pub fn hashmap() -> HashMap<String, Room> {
+        js_unwrap!($js_inner)
+    }
+
+    /// Retrieve the string keys of this object.
+    pub fn keys() -> Vec<String> {
+        js_unwrap!(Object.keys(Game.rooms))
+    }
+
+    /// Retrieve all values in this object.
+    pub fn values() -> Vec<Room> {
+        js_unwrap_ref!(Object.values(Game.rooms))
+    }
+
+    /// Retrieve a specific value by key.
+    pub fn get(name: LocalRoomName) -> Option<Room> {
+        js_unwrap_ref!(Game.rooms[@{name}])
+    }
 }
 
 /// See [http://docs.screeps.com/api/#Game.spawns]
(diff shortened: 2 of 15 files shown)


src/constants.rs
```diff
@@ -125,7 +125,7 @@ js_deserializable!(ReturnCode);
 /// ```no_run
 /// use screeps::{find, game, Room};
 ///
-/// let room: Room = game::rooms::get("E23S55").unwrap();
+/// let room: Room = game::rooms::get("E23S55".parse().unwrap()).unwrap();
```
ASalvail: There's a missing `)` before the `.parse()`.

src/game.rs
```diff
@@ -45,7 +45,29 @@ pub mod flags {
...
 /// [http://docs.screeps.com/api/#Game.rooms]: http://docs.screeps.com/api/#Game.rooms
 pub mod rooms {
-    game_map_access!(objects::Room, Game.rooms);
+    use std::collections::HashMap;
+
+    use crate::{local::RoomName, macros::*, objects::Room};
+
+    /// Retrieve the full `HashMap<RoomName, Room>`.
+    pub fn hashmap() -> HashMap<String, Room> {
+        js_unwrap!($js_inner)
```
ASalvail: That looks like a leftover from the macro call?

src/game.rs
```diff
@@ -45,7 +45,29 @@ pub mod flags {
 ///
 /// [http://docs.screeps.com/api/#Game.rooms]: http://docs.screeps.com/api/#Game.rooms
 pub mod rooms {
-    game_map_access!(objects::Room, Game.rooms);
+    use std::collections::HashMap;
+
+    use crate::{local::RoomName, macros::*, objects::Room};
+
+    /// Retrieve the full `HashMap<RoomName, Room>`.
+    pub fn hashmap() -> HashMap<String, Room> {
```
ASalvail: Isn't it a `HashMap<RoomName, Room>`?

src/constants.rs
daboross: I believe the current code is correct? We want to parse the string `"E23S55"`, and then send the result of that parse/unwrap operation into `rooms::get`.

src/game.rs
daboross: Right, thanks!

src/game.rs
daboross: It is - I think I just completely forgot to update the `hashmap` function.

src/constants.rs
ASalvail: Oh... I'm the one who missed the other parens. Nevermind me!

src/game.rs
daboross: Turns out there are some problems with getting a `HashMap<RoomName, Room>` directly from JS - stdweb only has conversions for `HashMap<String, V>`.

I've implemented it relatively inefficiently for now, and opened https://github.com/koute/stdweb/issues/359.

PR #221, opened 2019-08-12
Work around undefined hits and hitsMax

### LINKED ISSUE #139: Attackable structures don't always have hits early-game
For example, in tutorial 5, all of the extensions have `undefined` hits and `undefined` hitsMax.  
The Novice area's borders are Constructed walls but they don't have any hits either, because they are decaying.

Currently the API panics if I call `.hits()` on either of those things:
```
Panic error message: js_unwrap at 381 in {...}\screeps-game-api\src\objects\mod.rs: TypeMismatch { expected: "Number which fits into u32", actual: "undefined" }
```

I think these are the only 2 objects I've seen this happen with, but I'm not sure if there's any reasonable edge-case detection, other than appending a blanket `|| 0` for these 2 objects.

daboross:
Fixes #139.

This is a pretty minimal solution, not necessarily the most elegant one.

The main alternative I can think of would be to just do this for `StructureWall`, so that we properly handle indestructible novice / respawn area walls. We could just ignore the apparent problems in simulation mode.

However, I think node.js should be able to optimize this down so that it doesn't have that much of a practical effect.

### DIFF
--- a/src/objects.rs
+++ b/src/objects.rs
@@ -391,10 +391,22 @@ pub unsafe trait Withdrawable: RoomObjectProperties {}
 /// The reference returned from `AsRef<Reference>::as_ref` must be a valid
 /// target for `Creep.attack`.
 pub unsafe trait Attackable: RoomObjectProperties {
+    /// Retrieve this hits of this structure, or `0` if this structure doesn't
+    /// have a hit count.
+    ///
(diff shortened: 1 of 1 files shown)


daboross:
Documentation added! Hopefully this is clear enough.

PR #224, opened 2019-08-17
Update screeps-defaults.toml

ASalvail:
Add a comment for windows paths and add a branch option that is needed in `Copy`.

### DIFF
--- a/screeps-defaults.toml
+++ b/screeps-defaults.toml
@@ -11,7 +11,8 @@ branch = "default"
 # port = 443
 
 # [copy]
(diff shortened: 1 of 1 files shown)


screeps-defaults.toml
```diff
@@ -11,8 +11,9 @@ branch = "default"
 # port = 443
 
 # [copy]
-# destination = "your copy destination"
+# destination = "your copy destination"  # On Windows, use forward slashes instead of backslashes
```
daboross: I think a comment above the line could be slightly more consistent with others in the file, but it's not a big deal!

PR #229, opened 2019-08-19
Packed, typed object ids

### LINKED ISSUE #51: RoomObjectId type wrapper
TODO: bikeshed name

This would wrap the strings produced by `RoomObject.id` and functions like `get_object` would take some `T: Into<RoomObjectId>`.

At minimum to be useful I think the wrapper would need:

- `From<String>`, `From<str>`
- `Into<String>`
- `AsRef<str>`
- `Deref<str>`
- manual `Serialize`, `Deserialize` implementations which just serialize the underlying string (no wrapper showing in the serialized format)
- manual `TryFrom<Value>` implementation

---

One question is whether we want to make it generic over the type of thing it references. We could have `RoomObjectId<StructureSpawn>` vs. `RoomObjectId<StructureWall>` be different types, though I assume we wouldn't want to enforce that deserialized/created `RoomObjectId`s reference the right type.

### LINKED ISSUE #192: Store object ids as binary representation
As @Dessix discovered, object ids in screeps are always [mongo ObjectId](https://docs.mongodb.com/manual/reference/method/ObjectId/)s, or an [approximation with the same format](https://github.com/screeps/engine/blob/a4e431c9ee33dcc55c04f3adc479145c210c73fd/spec/helpers/mocks/common.js#L1-L6) on private servers. That is, they're always a 12-byte integer represented as a hex string.

Since 12 bytes is pretty huge and baked into `MongoDB`, we could consider it unlikely to ever change. If we want to, then, we could parse this string in methods dealing with IDs, and store IDs as a ~`[u8; 12]` internally~. `[32; 3]` would probably be nicer to work with and nicer for the optimizer than `[u8; 12]`.

Related: #51

daboross:
Resolves #51, resolve #192.

This is based on an initial implementation @Dessix made, but I removed some of the optimizations in favor of simpler code, and created a separate untyped version.

This implements the binary/packed 12-byte storage described in #192 for object ids, and implements fairly efficient transfers of these 12-byte values between rust and JS. All of the translations, formatting and parsing are unit tested for reliability (since there's no reliance on a screeps server for that)

This also implements a typed version of the id, `ObjectId<T>`. The type is purely for type inference and optional safety - it has a `into_type()` method to change the inner type arbitrarily, and can be freely transformed into and from untyped object ids. My thought was that allowing this kind of "free" transmutation would mean we can have serialization and deserialization which just works, and doesn't have to do anything like calling into JS and checking `Game.getObjectById()` to ensure it's the right type. It means we need that one technically uneccessary check if you get an ID from a thing and then immediately use `get_object_typed` during the same tick, but I think the advantages are worth it.

### DIFF
--- a/src/local/object_id/raw.rs
+++ b/src/local/object_id/raw.rs
@@ -198,6 +198,24 @@ impl RawObjectId {
     }
 }
 
+impl From<RawObjectId> for ArrayString<[u8; 24]> {
+    fn from(id: RawObjectId) -> Self {
+        id.to_array_string()
+    }
+}
+
+impl From<RawObjectId> for String {
+    fn from(id: RawObjectId) -> Self {
+        id.to_string()
+    }
+}
+
+impl From<[u32; 3]> for RawObjectId {
+    fn from(packed: [u32; 3]) -> Self {
+        Self::from_packed(packed)
+    }
+}
+
 #[cfg(test)]
 mod test {
     use super::RawObjectId;
--- a/src/local/object_id/raw.rs
+++ b/src/local/object_id/raw.rs
@@ -1,0 +1,260 @@
+use std::{
+    fmt::{self, Write},
+    str::FromStr,
+};
+
+use arrayvec::ArrayString;
+use serde::{Deserialize, Serialize};
+use stdweb::{Reference, UnsafeTypedArray};
+
+use super::errors::RawObjectIdParseError;
+use crate::{macros::*, traits::TryInto, ConversionError};
+
+const MAX_PACKED_VAL: u128 = 1 << (32 * 3);
+
+/// Represents an Object ID using a packed 12-byte representation
+///
+/// Each object id in screeps is represented by a Mongo GUID, which,
+/// while not guaranteed, is unlikely to change. This takes advantage of that by
+/// storing a packed representation.
+///
+/// To convert to a String in JavaScript, either use
+/// [`RawObjectId::to_array_string`], or [`RawObjectId::unsafe_as_uploaded`].
+/// See method documentation for more information.
+#[derive(Copy, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct RawObjectId {
+    packed: [u32; 3],
+}
+
+impl fmt::Debug for RawObjectId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("RawObjectId")
+            .field("packed", &self.packed)
+            .field("real", &self.to_array_string())
+            .finish()
+    }
+}
+
+impl fmt::Display for RawObjectId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut ints = self.non_zero_packed_ints().iter();
+        if let Some(first) = ints.next() {
+            write!(f, "{:x}", first)?;
+        }
+        for next in ints {
+            write!(f, "{:08x}", next)?;
+        }
+        Ok(())
+    }
+}
+
+impl FromStr for RawObjectId {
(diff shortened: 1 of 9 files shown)


src/local/object_id/raw.rs
```diff
@@ -0,0 +1,260 @@
...
+    /// })
+    /// .try_into()
+    /// .unwrap();
+    ///
+    /// let parsed = RawObjectId::from_packed_js_val(packed_obj_id).unwrap();
+    /// println!("found creep with id {}", parsed);
+    /// ```
+    pub fn from_packed_js_val(packed_val: Reference) -> Result<Self, ConversionError> {
+        let mut packed = [0u32; 3];
+        // TODO: make this more efficient, once we get mutable UnsafeTypedArrays.
```
ASalvail: Do you mind switching this to a tracking issue? We might never see it again otherwise :P 
(Feel free to leave the todo)

src/objects.rs
```diff
@@ -134,8 +134,33 @@ where
 
 /// Trait covering all objects with an id.
 pub unsafe trait HasId: RoomObjectProperties {
-    fn id(&self) -> String {
-        js_unwrap!(@{self.as_ref()}.id)
+    /// Retrieves this object's id as an untyped, packed value.
+    ///
+    /// This has no major differences from [`HasId::id`] besides the return
```
ASalvail: besides -> beside?

src/local/object_id/raw.rs
daboross: Will do!

src/objects.rs
daboross: Right - I think I almost never use `besides` correctly...

src/local/object_id/raw.rs
daboross: Opened https://github.com/rustyscreeps/screeps-game-api/issues/230.

src/objects.rs
ASalvail: Meh, I'm not married to either, hence the question mark. Feel free to leave it like that.

daboross:
With the other conversions - I'm happy to add  `Into<ArrayString<[u8; 24]>>`, `Into<String>` and `From<[u32; 3]>` - those seem natural.

I think `From<String>` would be a bit of a stretch though - conversion into `ObjectId` could definitely fail if it isn't in the right format, and we do have a `FromStr` implementation which allows using `string.parse()?`?

ASalvail:
You're right. Maybe an expensive `TryFrom<String>`, but that, I'd definitely leave for later.

daboross:
Sounds good!

I forgot you mentioned `Into<T>` too, and I didn't address that. What would you think of adding that as a method, like `try_resolve(&self) -> Result<Option<T>, ConversionError>` - maybe with a counterpart `resolve(&self) -> Option<T>` which panics on incorrect type?

Dessix:
I think a combination of `try_resolve` and `resolve` sounds good, but I may be biased, as- indeed- I did have that functionality in my own implementation.

There is also potential room for a threadlocal LRU which is per-tick reset caching objects by ID to references, reducing calls out to JS.

ASalvail:
Why not `TryInto`? I don't get why we don't rely on those here (except for the unsafe stuff) as it seems to be the most natural given the rest of the codebase.

Otherwise, I'm certainly not against `resolve` and `try_resolve`.

daboross:
I'd argue against it mostly because it doesn't feel like an `into` operation - an ID is a fairly separate bit of information from the object it names, and "into" conversions usually mean turning the same bit of data into another form? Like we're getting a lot more than another form of the information contained in `ObjectId`.

The other reason would be ease of use, since we would be implementing `TryInto<Option<T>>`.  `id.resolve()` seems like it would be nicer to use (and read) than `let x: Option<_> = id.try_into();` or `Option::try_from(id)`.

ASalvail:
Alright, you make a good case, I agree with you!

PR #231, opened 2019-08-21
Fix extra fixes / cleanups for object IDs

daboross:
Looking over the changes made in #229, I realized that using `u128` more could eliminate most/all of the manual formatting code for `RawObjectId`, and that a few more transitions which might be useful were missing.

This:
- adds impls `TryFrom<u128>`, `Into<u128>`, `From<[u32; 3]>`
- fixes the MAX_PACKED_VAL constant, previously off by 1
- removes manual formatting code in favor of calling using `u128`'s code

### DIFF
--- a/src/local/object_id/raw.rs
+++ b/src/local/object_id/raw.rs
@@ -10,7 +10,7 @@ use stdweb::{Reference, UnsafeTypedArray};
 use super::errors::RawObjectIdParseError;
 use crate::{macros::*, traits::TryInto, ConversionError};
 
-const MAX_PACKED_VAL: u128 = 1 << (32 * 3);
+const MAX_PACKED_VAL: u128 = (1 << (32 * 3)) - 1;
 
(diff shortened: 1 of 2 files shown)


src/local/object_id/raw.rs
```diff
@@ -250,6 +286,29 @@ mod test {
         }
     }
 
+    #[test]
```
ASalvail: Following @Dessix suggestion, it might be worth testing on wasm.

daboross: I believe CI _should_ be running all plain `#[test]` tests on WASM too as of https://github.com/rustyscreeps/screeps-game-api/commit/a1d21625896a8b567b1e05f623062d858b320788.

It doesn't seem to indicate what tests succeed, but a cursory test shows that locally, `cargo web test --verbose --nodejs` will fail if I insert a manual panic in here.

PR #233, opened 2019-08-21
Implement game resources

daboross:
This provides a way to access intershard / account-bound resources, which is currently just subscription tokens. https://docs.screeps.com/api/#Game.resources should no longer be missing!

I don't believe we had a tracking issue associated with this. I've tested `keys`, `get` and `hashmap`.


daboross:
Not sure! I think I've avoided doing it for a while because the key type wasn't just `String` - and I don't think we had `IntershardResourceType` until #164. But with the change to make `game::rooms` have `RoomName` as a key, this seemed reasonable.

PR #235, opened 2019-08-21
Have internal macro syntax more closely resemble the rust code it produces

### LINKED ISSUE #28: Macro syntax could be made more consistent in internal macros
Right now, internal macros have a variety of syntaxes. They do different things, but they could be cleaned up and have possibly improved legibility.

CC https://github.com/daboross/screeps-in-rust-via-wasm/pull/19/commits/46c03a6e02e3a7ccb132723d819cdcb646ce2458

daboross:
This changes a number of macros which produce either functions, methods or types to have macro input syntax more closely resembling the code it produces.

I'm hoping this will help newcomers to the codebase understand what things do a bit more, without having to dig into macro internals. 

The new macro syntax should includes visibility of types/methods produced, requires specifying `fn` / `struct` in the syntax, and uses `=` rather than `,` for the separator between the ident of what's being created and special values used by the macro (like JS names, or types in typesafe_*_constants).

Combined with #234, this closes #28.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -252,32 +252,32 @@ pub mod find {
     }
 
     typesafe_find_constants! {
-        CREEPS, 101, Creep;
-        MY_CREEPS, 102, Creep;
-        HOSTILE_CREEPS, 103, Creep;
-        SOURCES_ACTIVE, 104, Source;
-        SOURCES, 105, Source;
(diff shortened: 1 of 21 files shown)


daboross:
Glad to! And thanks for double checking all this.

PR #237, opened 2019-08-22
Split constants file into modules, and split exported constants into modules

### LINKED ISSUE #88: Update constants

daboross:
This addresses the last item in and closes #88.

The first commit here splits `constants.rs` into multiple files based on type of constant, and the last commit adds a bunch of modules which organize the exported constants based on the area of the game they affect. Probably way too many distinct modules, we can work on making that fewer.

The middle two commits are fairly small fixes semi-related to the above.


src/constants.rs
```diff
@@ -59,1279 +59,157 @@
...
-    Nuker = 18,
+/// Re-export of all constants related to minerals.
+pub mod minerals {
+    pub use super::{
+        numbers::{
+            mineral_min_amount, LAB_BOOST_ENERGY, LAB_BOOST_MINERAL, LAB_ENERGY_CAPACITY,
+            LAB_MINERAL_CAPACITY, LAB_REACTION_AMOUNT, LAB_UNBOOST_ENERGY, LAB_UNBOOST_MINERAL,
+            MINERAL_DENSITY_CHANGE, MINERAL_RANDOM_FACTOR, MINERAL_REGEN_TIME,
+        },
+        small_enums::Density,
```
ASalvail: Maybe include `ResourceType` here as it mostly pertains to minerals.

src/constants.rs
```diff
@@ -59,1279 +59,157 @@
...
-        ResourceType::Zynthium => Some(35_000),
-        ResourceType::Utrium => Some(35_000),
-        ResourceType::Catalyst => Some(35_000),
-        _ => None,
-    }
+/// Re-export of all constants related to sources.
+pub mod source {
+    pub use super::numbers::{
+        ENERGY_REGEN_TIME, INVADERS_ENERGY_GOAL, SOURCE_ENERGY_CAPACITY,
+        SOURCE_ENERGY_KEEPER_CAPACITY, SOURCE_ENERGY_NEUTRAL_CAPACITY,
```
ASalvail: Might want to add `ENERGY_DECAY` here.

src/constants.rs
```diff
@@ -59,1279 +59,157 @@
...
-        Self::from_str(&s).map_err(|_| {
-            D::Error::invalid_value(
-                Unexpected::Str(&s),
-                &"a known constant string in RESOURCES_ALL",
-            )
-        })
-    }
+/// Re-export of all constants related to `Resource`s.
+pub mod resource {
+    pub use super::{numbers::ENERGY_DECAY, types::ResourceType};
```
ASalvail: Might be worth adding `InterShardResource` here.

daboross: I was thinking of `resource` mainly being about dropped resources / the `Resource` type. I think `IntershardResourceType`, while named similarly, is unrelated?

ASalvail: Only problem is that it doesn't appear anywhere. Although, I guess it's still in `constants`...

daboross:
Rebased on master, added `ResourceType` to `minerals` module, and added links to submodule documentation:
screenshot-20190822-192511

I like the `resources` module even though it's... a bit bare. But I can see it being unfair to have it not others like `link`...

With `portal`, I separated it out mostly since I was thinking of its nature as a non-player-created and non-destroyable structure. Most of the things in `structure` seem useful for creating and destroying structures, and portals usually require fairly unique consideration. Though I guess everything matters for pathfinding, and that might be an arbitrary split...

src/constants.rs
daboross: I _believe_ that constant is more related to how dropped resources behave than `Source`s. Unless you'd propose merging the two modules? I can get behind that.

ASalvail: Well, beside fights, I expect to care about dropped energy when dealing with sources... Just a thought.

daboross: Ah, that's a good point. I'll stick it in here too!

ASalvail:
@Dessix @PatrickVdWillik any thoughts? It's meant as a convenience.

ASalvail:
Btw, it looks very good.  We're starting to have documentation that doesn't suck 🎉

src/constants.rs
daboross: true, it doesn't really have another.. place. I mostly figured it'd be useful alone, like `Terrain` or `Direction`. Maybe all of those should have some place too, though? I'm not sure.

Issue #226, opened 2019-08-18
Thought: implement PartialOrd/Ord for things?


daboross:
Right now, almost nothing in the library has a "natural" ordering, and thus nothing implements `Ord`.

While this is reasonable, it means nothing can be stored as a key in a `BTreeSet` or `BTreeMap`, which might be desireable.

@ASalvail What do you think of having `Position`, `RoomName`, `Resource`, and all other constants which implement `Hash` also implement `PartialOrd` and `Ord`, using an arbitrary but consistent-within-releases ordering?

ASalvail:
I'd say that's why `Hash` versions were invented: to give an unrelated mapping to integers that can be used for sorting? One could make an argument for `RoomName` or `Position`, but not `Resource`.

Any use cases for having an ordering on them that is all but arbitrary?
Maybe you have something @PatrickVdWillik ?

PatrickVdWillik:
Thanks for tagging me, I've been toying around with this myself but the biggest question I have is: Ordering in relation to what? Because while I got the stuff to work now, using a BTreeMap for my specific purpose, the calculation for the key value is offloaded into a separate function because the ordering has to make sense(in my case, a specific point on the map and the ordering is travel distance). 

If you're going to order based on an arbitrary property, how useful will the ordering be? There's a pretty select set of use cases for rooms ordered by travel distance from origin (0, 0) (for instance). Wouldn't a set of structs that allow custom ordering via a closure (or similar mechanism) perhaps be more useful?

ASalvail:
In that case, I'd vote to make sure `Hash` is derived, but let the end user figure out their own implementation for `PartialOrd`/`Ord`.

ASalvail:
I went through the structs that we define. The only one I can see where it would make sense to add `Ord` is with `screeps::constants::Density`.

daboross:
> I'd say that's why Hash versions were invented: to give an unrelated mapping to integers that can be used for sorting? One could make an argument for RoomName or Position, but not Resource.

`Hash` is definitely good for this. My only problem with it is how it doesn't _really_ integrate well with  things expecting something naturally ordered, like [`[T]::binary_search`](https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search) or BTreeMap. It's possible to use together, but it requires coming up with a hashing algorithm since `Hash` is agnostic to how the combination is done, and then writing that glue code to turn the `Hash` impl into an `Ord` impl.

[...]

PR #238, opened 2019-08-23
Implement Ord for Position, RoomName, ObjectId and RawObjectId

daboross:
This implements half of #226! `ObjectId` and `RawObjectId` are per the specification from my last comment, and `Position` / `RoomName` sort by `y` first rather than by `x` first.

### DIFF
--- a/src/local/room_name.rs
+++ b/src/local/room_name.rs
@@ -23,6 +23,8 @@ pub struct RoomName {
     ///
     /// This is the same representation of the upper 16 bits of [`Position`]'s
     /// packed representation.
+    ///
+    /// [`Position`]: crate::local::Position
     packed: u16,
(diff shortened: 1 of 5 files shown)


src/local/room_name.rs
```diff
@@ -23,6 +40,8 @@ pub struct RoomName {
     ///
     /// This is the same representation of the upper 16 bits of [`Position`]'s
     /// packed representation.
+    ///
+    /// [`Position`]: crate::local::Position
```
ASalvail: Oh, great thanks, I saw the doc compiler complaining about that earlier!

daboross: :+1: I almost made a separate PR but decided just to include it in this one

PR #242, opened 2019-08-26
Document `Position`'s behavior more thoroughly

### LINKED ISSUE #225: Document Position methods working locally and with world coordinates
Most of the methods on Position which can be are implemented in Rust rather than calling into JS. This fact isn't documented for most of them, though.

The other related, but separate issue is that most Position methods work on world coordinates, and this fact is not documented for many of them. get_range_to, for instance, seamlessly works on positions from different rooms - whereas the corresponding function in the RoomPosition JavaScript API returns Infinity for positions in different rooms.

daboross:
Fixes #225.

Adds documentation for various behaviors and functionalities which aren't obvious from the interface. There's still more to be done here, but I think this should help considerably.

### DIFF
--- a/src/local/room_position.rs
+++ b/src/local/room_position.rs
@@ -78,6 +78,27 @@ mod world_utils;
 ///   };
 ///   ```
 ///
+/// # Deserialization
+///
+/// `Position` implements `TryFrom<Value>`, allowing conversion from values
+/// retrieved from JavaScript. The implementation is fairly lenient, and will
+/// try to accept the value as any of the following things, in order:
+///
+/// - an integer representing the packedPos
+///   - this can be produced by retrieving the `__packedPos` field of a
+///     `RoomPosition`
+/// - an object with a `__packedPos` property
+///   - this allows converting from a JavaScript `RoomPosition` to a `Position`
+///     without referencing `__packedPos` manually, but is less efficient since
+///     it requires an extra callback into JavaScript to grab that field from
+///     within the conversion code
+/// - an object with `x`, `y` and `roomName` properties
+///   - this is mainly intended to decode `Position`s which were previously sent
+///     to JavaScript using `@{}` in `js!{}`, or serialized using
+///     [`serde::Serialize`]
+///   - this will also understand `RoomPosition`s in private servers versions
+///     `3.2.1` and below, prior to when `__packedPos` was added
+///
 /// # Ordering
 ///
 /// To facilitate use as a key in a [`BTreeMap`] or other similar data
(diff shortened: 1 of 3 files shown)


src/local/room_position.rs
```diff
@@ -78,6 +78,70 @@ mod world_utils;
...
+/// - an object with a `__packedPos` property
+///   - this allows converting from a JavaScript `RoomPosition` to a `Position`
+///     without referencing `__packedPos` manually, but is less efficient since
+///     it requires an extra callback into JavaScript to grab that field from
+///     within the conversion code
+/// - an object with `x`, `y` and `roomName` properties
+///   - this is mainly intended to decode `Position`s which were previously sent
+///     to JavaScript using `@{}` in `js!{}`, or serialized using
+///     [`serde::Serialize`]
+///   - this will also understand `RoomPosition`s in private servers versions
```
ASalvail: So... Saturday's problem shouldn't have happened?

src/local/room_position.rs
```diff
@@ -78,6 +78,70 @@ mod world_utils;
...
+/// let pos1 = Position::new(0, 0, "E1N1".parse().unwrap());
+/// let pos2 = Position::new(40, 20, "E1N1".parse().unwrap());
+/// assert_eq!(pos1 + (40, 20), pos2);
+/// ```
+///
+/// And the `Sub` implementation can be used to get the offset between two
+/// positions:
+///
+/// ```
+/// # use screeps::Position;
```
ASalvail: Maybe have an example where the resulting offset is negative?

src/local/room_position.rs
```diff
@@ -78,6 +78,70 @@ mod world_utils;
...
+///
+/// While this corresponds with the JavaScript `RoomPosition` type, it is not
+/// identical. In particular, all "calculation" methods which take in another
+/// position are re-implemented in pure Rust code, and behave slightly
+/// different.
+///
+/// For instance, [`Position::get_range_to`] operates on "world positions", and
+/// will return accurate distances for positions in different rooms. This is in
+/// contrast to `RoomPosition.getRangeTo` in JavaScript, which will return
+/// `Infinity` for positions from different rooms.
```
ASalvail: Might be worth adding where is positive/negative, and the range of numbers.

src/local/room_position/game_math.rs
```diff
@@ -48,6 +60,12 @@ impl Position {
         dx.abs().max(dy.abs()) as u32
     }
 
+    /// Checks whether this position is in the given range of another position.
+    ///
+    /// This operates on positions as "world positions", and may return true for
+    /// positions in different rooms which are still within the given range.
+    /// Note that the corresponding JavaScript method, `RoomPosition.inRangeTo`,
+    /// will always return `false` for positions from different rooms.
```
ASalvail: Should we implement that behavior in addition to what we have?

daboross:
Seems like something was up with travis's cache for the nightly build. Cleared those and restarted it.

PR #243, opened 2019-09-02
Some Position utility methods for approximate offsets

daboross:
Just some methods which could be useful?

I'm on the edge for whether these do belong in this library, but if they're useful to people they shouldn't hurt _too much_.

If not, maybe now would be a good time to add a "position_utils" addon library?

### DIFF
--- a/src/local/room_position/approximate_offsets.rs
+++ b/src/local/room_position/approximate_offsets.rs
@@ -1,0 +1,147 @@
+//! Methods related to approximating positions between other positions.
+use crate::objects::HasPosition;
+
+use super::Position;
+
+impl Position {
+    /// Calculates an approximate midpoint between this point and the target.
+    ///
+    /// In case of a tie, rounds towards this point.
+    ///
+    /// If `distance_towards_target` is bigger than the distance to the target,
+    /// the target is returned.
+    pub fn towards<T>(self, target: &T, distance_towards_target: i32) -> Position
+    where
+        T: ?Sized + HasPosition,
+    {
+        let target = target.pos();
+
+        let (offset_x, offset_y) = target - self;
+        let total_distance = offset_x.abs().max(offset_y.abs());
+        if distance_towards_target > total_distance {
+            return target;
+        }
+
+        let new_offset_x = (offset_x * distance_towards_target) / total_distance;
+        let new_offset_y = (offset_y * distance_towards_target) / total_distance;
+
+        self + (new_offset_x, new_offset_y)
+    }
+
+    /// Calculates an approximate midpoint between this point and the target.
+    ///
+    /// In case of a tie, rounds towards the target.
+    ///
+    /// If `distance_from_target` is bigger than the distance to the target,
+    /// this position is returned.
+    pub fn between<T>(self, target: &T, distance_from_target: i32) -> Position
+    where
+        T: ?Sized + HasPosition,
+    {
+        target.pos().towards(&self, distance_from_target)
+    }
+
+    /// Calculates an approximate midpoint between this point and the target.
+    ///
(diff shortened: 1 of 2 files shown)


src/local/room_position/approximate_offsets.rs
```diff
@@ -0,0 +1,147 @@
+//! Methods related to approximating positions between other positions.
+use crate::objects::HasPosition;
+
+use super::Position;
+
+impl Position {
+    /// Calculates an approximate midpoint between this point and the target.
```
ASalvail: This is not a midpoint (as in, it's not in the middle). It's `distance_towards_target` from this point towards the other point. (with big caveats)

ASalvail: Also, this line is suspiciously like the first one of the next function 😛

src/local/room_position/approximate_offsets.rs
```diff
@@ -0,0 +1,147 @@
...
+        self + (new_offset_x, new_offset_y)
+    }
+
+    /// Calculates an approximate midpoint between this point and the target.
+    ///
+    /// In case of a tie, rounds towards the target.
+    ///
+    /// If `distance_from_target` is bigger than the distance to the target,
+    /// this position is returned.
+    pub fn between<T>(self, target: &T, distance_from_target: i32) -> Position
```
ASalvail: Given that the other is `toward`, wouldn't this be more understandable as `from`?

src/local/room_position/approximate_offsets.rs
```diff
@@ -0,0 +1,147 @@
...
+    where
+        T: ?Sized + HasPosition,
+    {
+        target.pos().towards(&self, distance_from_target)
+    }
+
+    /// Calculates an approximate midpoint between this point and the target.
+    ///
+    /// In case of a tie, rounds towards the target.
+    pub fn midpoint_between<T>(self, target: &T) -> Position
```
ASalvail: suggestion: `midpoint_between` -> `midpoint_to`
It reads better.

src/local/room_position/approximate_offsets.rs
ASalvail: Maybe: `midpoint` -> `offset`

src/local/room_position/approximate_offsets.rs
```diff
@@ -0,0 +1,147 @@
...
+            return target;
+        }
+
+        let new_offset_x = (offset_x * distance_towards_target) / total_distance;
+        let new_offset_y = (offset_y * distance_towards_target) / total_distance;
+
+        self + (new_offset_x, new_offset_y)
+    }
+
+    /// Calculates an approximate midpoint between this point and the target.
```
ASalvail: Same: ~midpoint~ offset

src/local/room_position/approximate_offsets.rs
```diff
@@ -0,0 +1,147 @@
...
+    use super::Position;
+    use crate::RoomName;
+
+    fn test_rooms() -> impl Iterator<Item = RoomName> {
+        ["E0N0", "E20N20", "W20N0", "E20S20", "W20S20"]
+            .iter()
+            .map(|s| s.parse().unwrap())
+    }
+
+    fn pos(room: RoomName, x: u32, y: u32) -> Position {
```
ASalvail: `use super::Position::new as pos;`? 😛

daboross:
Sorry I've just kind of left this. Really glad to have your review, and once I have the time to be more active again, I'll be fixing the things! (and possibly just sending the code into a separate crate anyways...)

ASalvail (approved):

ASalvail:
Just for cleanup's sake.

PR #247, opened 2019-09-15
Made boost presence optional on Creep.body() calls

### LINKED ISSUE #245: Creep::body fails when parts aren't boosted
This line:

```rust
            let boost_v =
                js!(return __resource_type_str_to_num(@{self.as_ref()}.body[@{i}].boost););
```

Will throw an error, since `__resource_type_str_to_num` always requires valid input.

Dessix:
Fixes #245.

### DIFF
--- a/src/objects/impls/creep.rs
+++ b/src/objects/impls/creep.rs
@@ -27,7 +27,7 @@ impl Creep {
 
         for i in 0..len {
             let boost_v =
-                js!(return __resource_type_str_to_num(@{self.as_ref()}.body[@{i}].boost););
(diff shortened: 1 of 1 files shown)


src/objects/impls/creep.rs
```diff
@@ -27,7 +27,7 @@ impl Creep {
 
         for i in 0..len {
             let boost_v =
-                js!(return __resource_type_str_to_num(@{self.as_ref()}.body[@{i}].boost););
+                js!(const b=@{self.as_ref()}.body[@{i}].boost||null;return b&&__resource_type_str_to_num(b););
```
ASalvail: I'm nitpicking, but the line is fairly long...

PR #250, opened 2019-10-13
Change Position::in_range_to to be inclusive

j-browne:
This change makes the rust api consistent with the javascript api and
also with the implementation of Position::is_near_to. In the screeps
docs, it says that `isNearTo(target) == inRangeTo(target, 1)`. This now
satisfies that.

### DIFF
--- a/src/constants/look.rs
+++ b/src/constants/look.rs
@@ -19,7 +19,6 @@ use stdweb::Value;
 
 use super::Terrain;
 use crate::{
-    macros::*,
     objects::{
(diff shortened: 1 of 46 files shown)


ASalvail:
There is something wrong with the build, so I may take more time to merge this than intended. Sorry about that.

j-browne:
> There is something wrong with the build, so I may take more time to merge this than intended. Sorry about that.

It seems the breakage was caused by a change in `stdweb`. I fixed it by changing the dependency to `stdweb = "=0.4.19"`. Sorry, I'm on my phone, so I don't have it in front of me to check if that was right, but I hope it helps.

ASalvail:
Can you rebase off of #252 ?

PR #251, opened 2019-10-16
Screeps 4.x new objects: factories, invader cores, deposits, and ruins

shanemadden:
Tested relatively lightly on a private server, haven't used all parts of the new objects' APIs yet so further testing is needed, but basic interaction with each type has been successful, so it seems stable enough to review!


src/objects/impls/ruin.rs
```diff
@@ -0,0 +1,11 @@
+use crate::{
+    macros::*,
+    objects::{Ruin, Structure},
+};
+
+simple_accessors! {
+    impl Ruin {
+        pub fn structure() -> Structure = structure;
```
shanemadden: In further testing, calling this panics - I think due to needing to convert with `__structure_type_str_to_num`, but I could use an assist in how to hook this in correctly.

src/objects.rs
```diff
@@ -565,7 +603,13 @@ unsafe impl OwnedStructureProperties for StructureTower {}
 // NOTE: keep impls for Structure* in sync with accessor methods in
 // src/objects/structure.rs
 
+unsafe impl HasStore for Ruin {
+    fn store_capacity(&self) -> u32 {
+        0 // no storeCapacity property
```
ASalvail: That isn't true, unless you mean that we can't add to the storage?

src/objects.rs
```diff
@@ -408,6 +420,15 @@ pub unsafe trait Transferable: RoomObjectProperties {}
...
 pub unsafe trait Withdrawable: RoomObjectProperties {}
 
+/// Trait for all wrappers over Screeps JavaScript objects which can be the
+/// target of `Creep.harvest`.
+///
+/// # Contracts
+///
+/// The reference returned from `AsRef<Reference>::as_ref` must be a valid
+/// target for `Creep.harvest`.
+pub unsafe trait Harvestable: RoomObjectProperties {}
```
ASalvail: Neat!

src/objects.rs
```diff
@@ -603,6 +649,7 @@ unsafe impl HasCooldown for StructureTerminal {}
 // NOTE: keep impls for Structure* in sync with accessor methods in
 // src/objects/structure.rs
 
+unsafe impl CanDecay for Ruin {}
```
ASalvail: And `Deposit`?

src/objects/impls/deposit.rs
```diff
@@ -0,0 +1,14 @@
+use crate::{constants::ResourceType, macros::*, objects::Deposit};
+
+simple_accessors! {
+    impl Deposit {
+        pub fn last_cooldown() -> u32 = lastCooldown;
+        pub fn ticks_to_decay() -> u32 = ticksToDecay;
```
ASalvail: Usually handled by `CanDecay` trait.

src/objects.rs
```diff
@@ -167,16 +175,20 @@ pub unsafe trait HasId: RoomObjectProperties {
 impl_has_id! {
     ConstructionSite,
     Creep,
+    Deposit,
```
ASalvail: `Deposit` doesn't have an id in the API. Though it has one in the game... Interesting.

src/objects/impls/ruin.rs
ASalvail: The way `Structure` is built, it _shouldn't_ panic when used this way. @daboross any idea?
The `simple_accessor` calls `js_unwrap` which in turns use the `TryFrom` definition. My only guess would be that `TryFrom` returns `ConversionError`, 
In turn, it might be because those structure are "vanilla", ie without a structure type in the list we have?

@shanemadden do you have a panic trace?

src/objects.rs
shanemadden: Yeah, that's right - they're like the structure version of tombstones, they may have resources inside them when they're created but you can only ever withdraw.

src/objects/impls/deposit.rs
shanemadden: Good call - I duplicated that implementation from `Tombstone` so maybe it can be removed from there as well?

shanemadden:
Updated to cover most of your comments - thanks again!

The way those base commodities are ordered now will keep compatibility, but won't match `constants.js` - I added comments to hopefully clarify that for anyone doing a side-by-side comparison of the two files in the future, but they'd match in the documentation as well if they were implemented higher up (but out of numeric order) - let me know if you prefer that way.

src/objects/impls/ruin.rs
ASalvail: Seems like the `structure` method unwraps an `Err`. I'm really not sure why though.

ASalvail:
The problem is in one of those two functions of `structure.rs`. Can you look at some debug prints of these? I'm thinking that maybe the js `structureType` might be `undefined`.

```rust
fn get_structure_type(structure: &Reference) -> Result<StructureType, ConversionError> {
    (js! {
        return __structure_type_str_to_num(@{structure}.structureType);
    })
    .try_into()
    .map_err(Into::into)
}

impl FromExpectedType<Reference> for Structure {
    fn from_expected_type(reference: Reference) -> Result<Self, ConversionError> {
        let structure_type = get_structure_type(&reference)?;
        let structure = construct_structure_variants!(
            structure_type => reference.into_expected_type()?
        );

        Ok(structure)
    }
}
```

ASalvail:
What's the structure ruin you are testing this with?

shanemadden:
In this case it was an extension structure - easiest way to test has been to just destroy a full extension :)

ASalvail:
... btw, next time you have such a big PR, wait until the end before you merge the commits. It's much easier to review incrementally rather than go through the whole thing again. 😨

shanemadden:
Will do, sorry about that!

shanemadden:
Yup, the rebase to master is done - that's why I was fidding around with the rebase and just figured I'd squash at the same time (whoops).

Debug-printing the problem `.structure` out in the js conversion looks like this:

    {"id":"4195f3e48b64207","hits":0,"hitsMax":1000,"structureType":"extension","owner":{"username":"asd"},"my":true}

I haven't actually had to debug print from the library itself before so I'm not quite sure how to log from the `try_from` function, if you have advice there?

ASalvail:
just use js to `Console.log` to it. See the implementation of `log` in the starter script (no need for the whole `fern` stuff, just the logging).

shanemadden:
Per slack discussion, implemented a `structure_info()` function on ruins instead of `structure()` that returns a tuple of the structure type, id, and owner of the destroyed structure - instead of trying to get the limited object returned by `Tombstone.structure` to convert into a full `Structure`. I'm not a huge fan of this implementation but it's functional.

shanemadden:
Will do, thanks!

src/objects.rs
```diff
@@ -371,7 +383,7 @@ pub unsafe trait HasEnergyForSpawn: CanStoreEnergy {}
 ///
 /// The reference returned from `AsRef<Reference>::as_ref` must be have a
 /// `cooldown` properties.
-pub unsafe trait HasCooldown: StructureProperties {
+pub unsafe trait HasCooldown: RoomObjectProperties {
```
daboross: I think this is _technically_ a breaking change, so it probably wants a CHANGELOG entry?

shanemadden: Good point, done!

PR #252, opened 2019-10-17
Fix build - switch back to 2015-style macro_use imports for stdweb

daboross:
I think this is a good idea mostly because of stdweb's macros changing
internally, and them not having full 2018-style-supported macro usage.


ASalvail:
Thanks for going through the files and change it back.

daboross:
No problem! Hopefully this commit will be easy to revert in one go too once we can.

PR #258, opened 2019-10-23
Power Creeps and Effects

### LINKED ISSUE #158: Implement changes from the power creep era update
Here is the changelog:
https://blog.screeps.com/2019/03/power-era/

- [ ] Interface to `PowerCreep`
- [ ] Interface to `Game.powerCreeps` and `Game.gpl`
- [ ] Add `RoomObject.effects`
- [ ] Add `StructureController.isPowerEnabled`
- [x] Add interface to `RawMemory.interShardSegment`
- [x] We might have to deal with `RoomPosition` not returning strings anymore (see changelog)

### LINKED ISSUE #246: creep.id() panics if the creep is still spawning
Getting the creep via it's name seems to work. I guess the id isn't assigned until it actually spawns.

shanemadden:
Added power creeps and the necessary game and power spawn object interfaces to work with them, as well as effects on room objects which are used by both power creeps and strongholds.

Effects can have types representing powers or 'natural effects' (which only strongholds use currently) - this is set up as a nested enum to keep `PowerType` distinct from `EffectType`, though these could also just be a single enum with the relevant integer representations of both power and natural effects if that's overcomplicating things.

Implementations on creeps which are identical to those on power creeps have been moved to a `SharedCreepProperties` trait which both creeps and power creeps use - I'm not a huge fan of this naming if there's other ideas.

Several attributes on power creeps not spawned in the world are undefined in the same way as the issue reported in #246, causing a number of potential panics.  This switches just the directly-impacted functions to return a `Result`, which might be a start toward #3 - currently just bubbling up serde's `ConversionError`s which seems potentially not ideal? If this ends up being the final implementation it's probably a good time/release to make similar result changes in other places.

Resolves #158

### DIFF
--- a/src/constants/numbers.rs
+++ b/src/constants/numbers.rs
@@ -276,8 +276,7 @@ pub const INVADERS_ENERGY_GOAL: u32 = 100000;
 
 pub const SYSTEM_USERNAME: &str = "Screeps";
 
-pub const SIGN_PLANNED_AREA: &str =
-    "A new Novice or Respawn Area is being planned somewhere \
+pub const SIGN_PLANNED_AREA: &str = "A new Novice or Respawn Area is being planned somewhere \
      in this sector. Please make sure all important rooms are reserved.";
 
 // EVENT_* constants in src/objects/impls/room.rs
--- a/src/constants/types.rs
+++ b/src/constants/types.rs
@@ -536,7 +536,7 @@ pub enum PowerType {
     OperateStorage = 4,
     OperateLab = 5,
     OperateExtension = 6,
-    OperateObserve = 7,
+    OperateObserver = 7,
     OperateTerminal = 8,
     DisruptSpawn = 9,
     DisruptTower = 10,
@@ -552,7 +552,7 @@ pub enum PowerType {
 
 js_deserializable!(PowerType);
 
-/// Translates effect types which includes both `POWER_*` and `EFFECT_*` constants.
+/// Translates effect types which includes both `PWR_*` and `EFFECT_*` constants.
 #[derive(
     Copy, Clone, Debug, PartialEq, Eq, Hash, FromPrimitive, Serialize_repr, Deserialize_repr,
 )]
@@ -583,7 +583,7 @@ impl<'de> Deserialize<'de> for EffectType {
             4 => EffectType::PowerEffect(PowerType::OperateStorage),
             5 => EffectType::PowerEffect(PowerType::OperateLab),
             6 => EffectType::PowerEffect(PowerType::OperateExtension),
-            7 => EffectType::PowerEffect(PowerType::OperateObserve),
+            7 => EffectType::PowerEffect(PowerType::OperateObserver),
             8 => EffectType::PowerEffect(PowerType::OperateTerminal),
             9 => EffectType::PowerEffect(PowerType::DisruptSpawn),
             10 => EffectType::PowerEffect(PowerType::DisruptTower),
@@ -600,7 +600,7 @@ impl<'de> Deserialize<'de> for EffectType {
             _ => {
                 return Err(D::Error::invalid_value(
                     Unexpected::Unsigned(effect as u64),
-                    &"a valid POWER_* or EFFECT_* type integer",
+                    &"a valid PWR_* or EFFECT_* type integer",
                 ))
             }
         };
--- a/src/constants/types.rs
+++ b/src/constants/types.rs
@@ -513,7 +513,7 @@ impl ResourceType {
 
 js_deserializable!(ResourceType);
 
-/// Translates the `POWER_CLASS` constants.
+/// Translates the `POWER_CLASS` constants, which are classes of power creeps
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize_repr, Deserialize_repr, FromStr)]
 #[repr(u8)]
 pub enum PowerCreepClass {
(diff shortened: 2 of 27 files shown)


src/objects.rs
```diff
@@ -120,12 +122,12 @@ reference_wrappers! {
 ///
 /// This can be freely implemented for anything with a way to get a position.
 pub trait HasPosition {
-    fn pos(&self) -> Position;
+    fn pos(&self) -> Result<Position, ConversionError>;
```
daboross: What would you think of just not implementing `HasPosition` and other traits for `PowerCreep` and instead requiring first converting to an "InstantiatedPowerCreep" or similar before being able to call relevant methods?

I'd just suggest it because changing the traits like this requires everyone to deal with usually-non-existent error cases dealing with every other object.

The other disadvantage would be that this makes methods like `.get_range_to` panic when called with a power creep, something that "seems" valid. Ideally we should be able to differentiate between things which can be put as an argument to that and not at compile time, with traits or different types, rather than erroring out at runtime.

Thoughts?

shanemadden: Agreed, this kind of change to these methods makes them a lot more awkward to use - my initial attempt at implementation used an enum to wrap an alive-or-dead-power-creep, but that ended up being a mess - this sounds better ;)

The comparison I'm thinking of is that we'd have something akin to a `std::rc::Weak` upgrade into a `std::rc::Rc` - taking the maybe-alive-power-creep reference object reference and upgrading it to get a we-know-it's-alive object with room object properties, `HasId`, `Attackable` etc.

Because this is only applicable to the results of accessing `game::power_creeps` (and `game::creeps` in the #246 case) case and not for objects found via `find()`/`look()`, my thinking is to have `PowerCreep` as the 'fully alive power creep' representation (which would be the find/look result, intuitively) and to add `PowerCreepInfo` or `GamePowerCreep` or somesuch which would be the version of the reference that can be `upgrade()`-ed to a full `PowerCreep` if it's spawned on the current shard.

I could use some guidance on how to get this implemented - I can't quite get my head around how to get that 'upgrade()' function set up to get the underlying js ref moved from `ShrodingersPowerCreep` to `PowerCreep` during the `upgrade()`, to wrap it in the 'real' object type.

shanemadden: Worked it out and went with `AccountPowerCreep`, should be a much less breaking changeset now :)

daboross: Awesome!

> The comparison I'm thinking of is that we'd have something akin to a std::rc::Weak upgrade into a std::rc::Rc - taking the maybe-alive-power-creep object reference and upgrading it to get a we-know-it's-alive object with room object properties, HasId, Attackable etc.

This sounds like a sane comparison to me. The new scheme looks good!

shanemadden:
Ok, last implementation question - should I add a wrapper around `game::creeps::*` returns to account for the potentially missing id in #246 in the same way `AccountPowerCreep` does for power creeps or leave fixing that for another time?

src/objects.rs
ASalvail: > `ShrodingersPowerCreep`

🤣 

I think I like it better than `AccountPowerCreep` 😛

ASalvail:
> Ok, last implementation question - should I add a wrapper around `game::creeps::*` returns to account for the potentially missing id in #246 in the same way `AccountPowerCreep` does for power creeps or leave fixing that for another time?

This one's big enough already. I suggest keeping it separate. PRs aren't expensive!

src/constants/types.rs
```diff
@@ -513,19 +513,30 @@ impl ResourceType {
...
+}
+
+js_deserializable!(PowerCreepClass);
+
+/// Translates the `PWR_*` constants, which are types of powers used by power creeps
 #[derive(
     Copy, Clone, Debug, PartialEq, Eq, Hash, FromPrimitive, Serialize_repr, Deserialize_repr,
 )]
-#[repr(u8)]
+#[repr(u16)]
```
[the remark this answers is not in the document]
shanemadden: I think my thinking was to make it match the size of `NaturalEffectType` but I suppose there isn't really any reason to keep those synced!

shanemadden:
@ASalvail @daboross Have a couple more minor changes on here that could use a quick eyeball before they get merged - thanks!

src/constants/numbers.rs
```diff
@@ -276,8 +276,7 @@ pub const INVADERS_ENERGY_GOAL: u32 = 100000;
 
 pub const SYSTEM_USERNAME: &str = "Screeps";
 
-pub const SIGN_PLANNED_AREA: &str =
-    "A new Novice or Respawn Area is being planned somewhere \
+pub const SIGN_PLANNED_AREA: &str = "A new Novice or Respawn Area is being planned somewhere \
```
ASalvail: rustfmt changed it's default value?

src/objects/impls/structure_tower.rs
```diff
@@ -1,18 +1,27 @@
...
-    pub fn attack(&self, target: &Creep) -> ReturnCode {
+    pub fn attack<T>(&self, target: &T) -> ReturnCode
+    where
+        T: Attackable,
+    {
         js_unwrap! { @{self.as_ref()}.attack( @{target.as_ref()} ) }
     }
 
-    pub fn heal(&self, target: &Creep) -> ReturnCode {
+    pub fn heal<T>(&self, target: &T) -> ReturnCode
```
ASalvail: Can you heal enemy creeps?

src/constants/numbers.rs
shanemadden: The joys of the nightly branch apparently!

src/objects/impls/structure_tower.rs
shanemadden: Yup, sure can!  You usually wouldn't want to, but there is a useful uncommon case in healing allies' creeps (though my intent for this change was just to make it consistent with the args for the heal function on `Creep`)

PR #260, opened 2019-10-26
Update market API for new features and compatibility with current Screeps

### LINKED ISSUE #256: Screeps 4.x market API updates
There are two changes to the market interface in 4.0:

- [ ] `Game.market.createOrder()` now takes an object instead of the parameters in order; the old form still works but is no longer documented and presumably deprecated.
- [ ] `Game.market.getHistory()` was added with historical price data.

shanemadden:
Resolves #256, I bit off a little more than I thought I had with this one.  It turned out that intershard resources had broken a lot of the market code's assumptions for deserializing the struct representations, so a bunch of breaking changes were needed to get these functions to where they wouldn't panic on the live servers.

Market API calls can return the string representations of either a resource or intershard resource, so I added an new enum wrapper to contain them in the places the market code accepts or returns a variable type; moving the integer representation of `token` out of the same spaces used for normal resources was needed in this case, so I used `1001` to kinda match how the new natural effects are offset by the base game, but opinions welcome if another value is preferred.

### DIFF
--- a/src/constants/types.rs
+++ b/src/constants/types.rs
@@ -118,8 +118,8 @@ js_deserializable!(StructureType);
 ///
 /// *Note:* This constant's `TryFrom<Value>`, `Serialize` and `Deserialize`
 /// implementations only operate on made-up integer constants. If you're ever
-/// using these impls manually, use the `__intershard_resource_type_num_to_str`
-/// and `__intershard_resource_type_str_to_num` JavaScript functions,
+/// using these impls manually, use the `__resource_type_num_to_str`
+/// and `__resource_type_str_to_num` JavaScript functions,
 /// [`FromStr`][std::str::FromStr] or
 /// [`IntershardResourceType::deserialize_from_str`].
 ///
@@ -522,11 +522,10 @@ pub enum MarketResourceType {
     IntershardResource(IntershardResourceType),
 }
 
-impl<'de> Deserialize<'de> for MarketResourceType {
-    fn deserialize<D>(d: D) -> Result<Self, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
+impl MarketResourceType {
+    /// Helper function for deserializing from a string rather than a fake
+    /// integer value.
+    pub fn deserialize_from_str<'de, D: Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
         let s: Cow<'de, str> = Cow::deserialize(d)?;
 
         ResourceType::from_str(&s)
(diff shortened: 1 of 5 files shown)


src/constants/types.rs
```diff
@@ -513,6 +514,50 @@ impl ResourceType {
...
+
+impl<'de> Deserialize<'de> for MarketResourceType {
+    fn deserialize<D>(d: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s: Cow<'de, str> = Cow::deserialize(d)?;
+
+        // first try to deserialize as a ResourceType
+        match ResourceType::from_str(&s) {
```
ASalvail: You could probably have this look a bit nicer using the `Result` methods, but that code aesthetic.

src/game.rs
```diff
@@ -65,14 +65,12 @@ pub mod resources {
 
     /// Retrieve the string keys of this object.
     pub fn keys() -> Vec<IntershardResourceType> {
-        js_unwrap!(Object
-            .keys(Game.resources)
-            .map(__intershard_resource_type_str_to_num))
+        js_unwrap!(Object.keys(Game.resources).map(__resource_type_str_to_num))
```
ASalvail: I'm surprised `rustfmt` changed its mind on that one.

src/game/market.rs
```diff
@@ -160,41 +181,101 @@ pub fn change_order_price(order_id: &str, new_price: f64) -> ReturnCode {
...
+            }
+        }
     }
 }
 
-pub fn deal(order_id: &str, amount: u32, target_room: &Room) -> ReturnCode {
-    js_unwrap! {Game.market.deal(@{order_id}, @{amount}, @{target_room.name()})}
+/// Execute a market trade
+///
+/// target_room is your owned room whose terminal will send or receive resources in this
```
ASalvail: target_room -> `target_room`

src/game.rs
shanemadden: I think this might be because I'm not on the nightly?

shanemadden: Ahh, shorter js function name pushed it back under the one line char limit.

src/constants/types.rs
shanemadden: Updated, I tried to make it not too big of a mess but the formatter had other ideas ;)

src/game/market.rs
```diff
@@ -160,41 +185,102 @@ pub fn change_order_price(order_id: &str, new_price: f64) -> ReturnCode {
...
 }
 
-pub fn deal(order_id: &str, amount: u32, target_room: &Room) -> ReturnCode {
-    js_unwrap! {Game.market.deal(@{order_id}, @{amount}, @{target_room.name()})}
+/// Execute a market trade
+///
+/// `target_room` is your owned room whose terminal will send or receive
+/// resources in this transaction, or `None` if this is an order for an
+/// intershard resource type
+pub fn deal(order_id: &str, amount: u32, target_room: Option<RoomName>) -> ReturnCode {
```
shanemadden: Actually, should this (and the others like it) be `Option<&RoomName>` instead?

daboross: Since `RoomName` is Copy, this should be fine. It's a very small struct (not bigger than `u32`), and practically free to copy around.

src/constants/types.rs
```diff
@@ -513,6 +514,46 @@ impl ResourceType {
...
+}
+
+impl Serialize for MarketResourceType {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match self {
+            MarketResourceType::Resource(ty) => ty.serialize(serializer),
+            MarketResourceType::IntershardResource(ty) => ty.serialize(serializer),
```
daboross: Are you sure this `Serialize` impl actually matches the `Deserialize` one?

As I recall, `ResourceType` and `IntershardResourceType` both implement `Serialize` / `Deserialize` by serializing to our fake numbers, not to string representations. Since the `Deserialize` impl for `MarketResourceType` depends on string representations, I think we probably want to serialize to strings as well here. Either that, or expand the `Deserialize` impl to accept the integers too?

src/game/market.rs
```diff
@@ -60,94 +59,120 @@ impl OrderType {
...
-    let order_book_value = js! {
-        return Game.market.orders;
-    };
-    order_book_value.try_into().unwrap()
+    js_unwrap!(Game.market.orders)
 }
 
-pub fn calc_transaction_cost(amount: u32, room1: &Room, room2: &Room) -> f64 {
-    js_unwrap!(Game.market.calcTransactionCost(@{amount}, @{room1.name()}, @{room2.name()}))
+pub fn calc_transaction_cost(amount: u32, room1: &RoomName, room2: &RoomName) -> f64 {
```
daboross: Since `RoomName` is the same size as a `u16`, and copy, it should be more efficient to take them by value here and below?

Basing this off the logic from the trivial-copy-pass-by-ref clippy lint: https://rust-lang.github.io/rust-clippy/master/#trivially_copy_pass_by_ref

src/constants/types.rs
```diff
@@ -125,10 +125,11 @@ js_deserializable!(StructureType);
 ///
```
daboross: If it wouldn't be a nuisance, could you update the doc-comment for `IntershardResourceType` to mention `__resource_type_str_to_num` rather than `__intershard_resource_type_str_to_num` to accompany the change?

src/constants/types.rs
shanemadden: Good catch - needed to be updated to have both the magic-int parse on the normal deserialize, as well as adding the `MarketResourceType::deserialize_from_str` function for processing API value returns.  I made it yet-another-exhaustive-match-of-the-whole-list-of-resources which is a little painful, but I couldn't figure out a more clever way to try deserializing as one type then the other; let me know if it should be improved!

ASalvail:
@daboross you want to have another look?

src/constants/types.rs
daboross: Looks good to me! There might be something we could do with trying to deserialize ResourceType from the  `u16` first, then if that fails trying MarketResourceType, but I'm not entirely sure if that would be easier/better or how possible it'd be.

Sorry about not getting back to this immediately!

PR #263, opened 2019-11-10
Change room() to returning Option<Room> to fix panics when undefined

shanemadden:
The game API documentation covers two expected cases where `room` on a room object will be undefined ([link](https://docs.screeps.com/api/#RoomObject.room)):

> The link to the Room object. May be undefined in case if an object is a flag or a construction site and is placed in a room that is not visible to you.

This changes the return type of `RoomObjectProperties::room()` from `Room` to `Option<Room>` to prevent panic in the cases where that room isn't visible and `undefined` is returned by js.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Unreleased
 - Remove explicitly implemented `Creep::energy()` function which used deprecated `.carry`, now
   using the `energy()` implementation from `HasStore`
 - Change `RoomObjectProperties::room()` to return `Option<Room>` to handle the cases that the base
-  game API leaves it undefined: for construction sites and flags in non-visible rooms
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ Unreleased
 - Remove `StructurePowerSpawn::power()` and `power_capacity()` (replaced with `HasStore` functions)
 - Remove explicitly implemented `Creep::energy()` function which used deprecated `.carry`, now
   using the `energy()` implementation from `HasStore`
+- Change `RoomObjectProperties::room()` to return `Option<Room>` to handle the cases that the base
+  game API leaves it undefined: for construction sites and flags in non-visible rooms
```
ASalvail: Breaking.

PR #267, opened 2019-12-28
Screeps 4.1 new features and deprecations

shanemadden:
Updates to track changes to the base game API in version 4.1 - note that this version is currently in the public test realm branch and is not yet released, so these endpoints don't work on non-PTR servers yet.  Changelog [here](https://screeps.com/forum/topic/2900/ptr-changelog-2019-12-21).

 - `Game.map.isRoomAvailable` is now giving deprecation warnings in the base game, marked as deprecated.  It's replaced with `Game.map.getRoomStatus`, which returns more details about the room's state.
 - Labs can now break down components; added `StructureLab::reverse_reaction`

### DIFF
--- a/src/game/map.rs
+++ b/src/game/map.rs
@@ -10,7 +10,6 @@ use serde::{
     de::{Deserializer, Error as _, Unexpected},
     Deserialize,
 };
-use serde_repr::{Deserialize_repr, Serialize_repr};
 use stdweb::Value;
 
 use crate::{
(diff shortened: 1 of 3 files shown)


src/game/map.rs
```diff
@@ -58,10 +63,52 @@ pub fn get_world_size() -> u32 {
 /// See [http://docs.screeps.com/api/#Game.map.isRoomAvailable]
 ///
 /// [http://docs.screeps.com/api/#Game.map.isRoomAvailable]: http://docs.screeps.com/api/#Game.map.isRoomAvailable
+#[deprecated(
```
ASalvail: No need for that. The point of being in ver <1.0 is to savagely break the API every 0.x update.

ASalvail: Ah, the changes are in PTR... Then I'm not even sure whether we should deprecate it right now. Sometimes, features spend months in PTR before getting released.

src/game/map.rs
```diff
@@ -58,10 +63,52 @@ pub fn get_world_size() -> u32 {
...
+/// [http://docs.screeps.com/api/#Game.map.getRoomStatus]: http://docs.screeps.com/api/#Game.map.getRoomStatus
+pub fn get_room_status(room_name: RoomName) -> MapRoomStatus {
+    js_unwrap!(Game.map.getRoomStatus(@{room_name}))
+}
+
+/// Represents the availability and respawn/novice state of a room on the map
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct MapRoomStatus {
+    #[serde(deserialize_with = "RoomStatus::deserialize_from_str")]
```
ASalvail: Not sure you need this. I believe enums can be deserialized from strings out of the box, if you don't use `repr(u8)`, which I think is unnecessary here.

shanemadden:
Updated, thanks!

Should be no changes needed on our end for the changes to the return codes. 👍 

These changes definitely shouldn't be in a released version until the changes are released on live, which probably means it's a good idea to wait to merge until after that happens.

PR #282, opened 2020-02-16
Fix has_owner

smessmer:
It's not the `my` property but the `owner` property that gets undefined when a structure is unowned.

Tested only for StructureController, which returns `controller.my == false` in JS for unowned controllers but I think this likely extends to other structures too.


shanemadden:
I had actually bumped into this same issue in my bot and had just worked around it by matching on whether `controller.owner_name()` returning something or nothing to determine if a controller is owned instead of using `has_owner` - and of course forgot to make a note of it :) Thanks!

PR #284, opened 2020-02-27
Filter hidden _sum key from store keys.

Azaril:
There is an undocumented "_sum" key on the store API that contains the total resources available on the store. This breaks querying the store types which attempts to map all the keys to a resource type.

I did not change the store_total API to use the hidden key as despite it being faster, it is not documented in the official API and may just be an implementation detail.

### DIFF
--- a/src/objects.rs
+++ b/src/objects.rs
@@ -346,7 +346,7 @@ pub unsafe trait HasStore: RoomObjectProperties {
     }
 
     fn store_types(&self) -> Vec<ResourceType> {
-        js_unwrap!(Object.keys(@{self.as_ref()}.store).map(__resource_type_str_to_num))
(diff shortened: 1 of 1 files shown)


Azaril:
I'd like to keep this on hold until I hear back from the screeps dev team. The _sum key actually causes issues causing double calculation of resources in the game code.

I have a thread open on the forum here: https://screeps.com/forum/topic/2926/getusedcapacity-fast-cache-causes-double-counting-of-resources

Azaril:
Closing this PR as the bug was reported to the dev team and fixed. (They also added significant testing coverage for the store APIs as it's a fairly quirky interface.)

ASalvail:
Thanks for the update!

PR #286, opened 2020-03-15
Handle negative ticksToRegeneration return values

shanemadden:
On the official servers, sources in rooms that regenerate while the room is considered 'inactive' by the servers can sometimes return negative values for `ticksToRegeneration` - this wraps the returned value in a `Math.max` with 0 to handle that case.  Since I'd just made that an `Option<u32>` in  #285 but that condition is also handled by this fix, I switched the returned value back to no longer be option-wrapped (and applied the change to minerals for consistency, even though I'm not certain they can also get negative values).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,8 +51,8 @@ Unreleased
 - Fix `OwnedStructureProperties::has_owner()` to correctly return false for unowned structures
 - Work around a case where `map::describe_exits()` would panic when a private server returns null
   for an unavailable room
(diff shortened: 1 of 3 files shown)


ASalvail:
But isn't 0 and a negative number two different states?

shanemadden:
In practice, I think they'll always match - the negative values occur only on a fully-charged source (which would normally be returning `undefined` from js in this case but due to server-side weirdness, seems to occasionally get into a state where it returns a negative number that keeps on decrementing every tick), so a 0 would now always represent something that's fully-regenerated and doesn't have a timer active, coalescing both the 'negative-integer' and 'undefined' flavors of that state.

PR #288, opened 2020-03-26
Disable nodejs test due to upstream cargo-web issues with rust >= 1.44

shanemadden:
Temporarily comments out the test that's failing on current nightly builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,4 +33,3 @@ script:
 jobs:
   allow_failures:
     - rust: nightly
-    - rust: stable
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,5 @@ script:
 - cargo web test --verbose --nodejs
 jobs:
   allow_failures:
-    - rust: 
-      - nightly
-      - stable
+    - rust: nightly
+    - rust: stable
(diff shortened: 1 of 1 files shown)


ASalvail:
Wouldn't it be easier to lock the api to rust 1.43?

shanemadden:
Quite possibly - not sure how to do that ;)

ASalvail:
Right, that'd be in the testing environment. @daboross thoughts?

daboross:
Both sound reasonable to me? If it's _just_ the test that's failing, though, and not the entire build, then this seems reasonable.

We could have the tests only run on 1.43.0, but I don't know if the complexity would be worth it.

I don't know why travis is just at "Expected", though, not success or failing...

shanemadden:
Re-opening since we're hard blocked on these failures on stable in addition to nightly, so the workaround in #301 is no longer doing the job; unsure if this'll build or if we need additional tweaks to the travis.yml.

I've got a pretty bare hello world working with bindgen that's passing strings successfully (but not references yet), so we've got a couple steps down that road - but unfortunately it'll be a long one, and we'll be tied to stdweb's toolchain for a while yet.

Azaril:
For what it's worth I have identified the core issue and reported it to cargo web - including specifics of potential fixes. However, both cargo-web and stdweb appear to be unmaintained with no updates in months. (I have PRs open against the projects that haven't been looked at.)

I'm unfortunately inclined to disable to tests for the moment so we can keep on latest stable. A migration to bindgen seems somewhat inevitable due to lack of support from cargo web/stdweb.

PR #289, opened 2020-03-26
0.8.0 release changelog prep

shanemadden:
Reorganize changelog to prep for a release, ~~as well as a minor fix to the visibility of the MapRoomStatus struct~~ (handled in #300)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,61 +1,76 @@
 Unreleased
 ==========
 
-- Add `game::map::get_room_status()` as interface to new `Game.map.getRoomStatus()` function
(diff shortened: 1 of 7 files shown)


ASalvail:
We'll need to go back over the last flurry of PRs, I don't think they all had changelogs additions.

shanemadden:
Yeah - I thought of and added 'em for #295 and #300, can definitely do a more through verification before merging this.

shanemadden:
Confirmed that we're all good - this covers the couple that didn't get changelogs, and added changelog entries for the most recent couple merges. 👍

PR #290, opened 2020-03-27
Adds find_closest_by_path() method to position.

jhessin:
Signed-off-by: Jim Hessin <jhessin@gmail.com>

I would really like to implement the vector version of this method, but I'm not familiar enough with the traits involved in the macros.


ASalvail:
It didn't end up working?

Issue #79, opened 2018-10-08
Creep.move_to* methods are missing the visualize_path_style option


ASalvail:
Related to #46 

Implementation of the `visualize_path_style` optional argument has been postponed until a decision is made about room visuals.

jhessin:
What decision needs to be made about room visuals?

Azaril:
Here's a PR for the full API: https://github.com/rustyscreeps/screeps-game-api/pull/306

ASalvail:
> What decision needs to be made about room visuals?

Mostly, how to represent it in rust. Given how long this has been open, Azaril's version, if it works (😛) will probably be the answer.

Azaril:
Closing as resolved with https://github.com/rustyscreeps/screeps-game-api/pull/306

Issue #46, opened 2018-09-17
What kind of RoomVisual support do we want?


daboross:
Opening this as a separate issue from #30 since there are multiple ways we could support this and discussing it seems fairly self contained.

`RoomVisual` allows drawing nice graphics on rooms. It's a convenient API. However, all of the JS code in `RoomVisual` does is create the right JSON structures then call `console.addVisual(room_name, structure)`.

In implementing bindings for `RoomVisual`, we could either wrap `RoomVisual`'s methods themselves, or just wrap `console.addVisual` and create a `RoomVisual` equivalent ourselves in pure rust code.

ASalvail:
The latter seems to make more sense. It'd greatly reduce the overhead of communicating through the screeps API.

I'm curious however: would we bypass [this restriction](https://docs.screeps.com/api/#RoomVisual)
> However, there is a usage limit: you cannot post more than 500 KB of serialized data per one room (see getSize method).

daboross:
I don't believe we would - the check is done in `addVisual` itself rather than in any of the wrapper methods: https://github.com/screeps/engine/blob/551aa41163c45273d706ce238c6b35f379c0149e/src/game/console.js#L40

Azaril:
Here's a PR for the full API: https://github.com/rustyscreeps/screeps-game-api/pull/306

Azaril:
Closing as resolved with https://github.com/rustyscreeps/screeps-game-api/pull/306

PR #292, opened 2020-04-13
Implement move_to visualization options

drewcrawford:
This implements the visualization options for move_to in #79.  

Introduces a few types that could be useful for someday implementing #46.  However a more comprehensive implementation of `RoomVisual` is outside the scope of this PR.

A lot of the JS serialization stuff is new to me, send feedback

### DIFF
--- a/src/objects/creep_shared.rs
+++ b/src/objects/creep_shared.rs
@@ -1,6 +1,7 @@
 use std::{marker::PhantomData, mem};
 
 use scoped_tls::scoped_thread_local;
+use serde::{Deserialize, Serialize};
 use stdweb::Reference;
 
 use crate::{
@@ -77,7 +78,7 @@ pub unsafe trait SharedCreepProperties: RoomObjectProperties {
             reuse_path,
             serialize_memory,
             no_path_finding,
-            // visualize_path_style,
+            visualize_path_style,
             find_options:
                 FindOptions {
                     ignore_creeps,
@@ -129,7 +130,6 @@ pub unsafe trait SharedCreepProperties: RoomObjectProperties {
         let callback_lifetime_erased: Box<
             dyn Fn(RoomName, Reference) -> Option<Reference> + 'static,
         > = unsafe { mem::transmute(callback_type_erased) };
-
         // Store callback_lifetime_erased in COST_CALLBACK for the duration of the
         // PathFinder call and make the call to PathFinder.
         //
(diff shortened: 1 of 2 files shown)


src/objects/creep_shared.rs
```diff
@@ -262,14 +262,79 @@ pub unsafe trait SharedCreepProperties: RoomObjectProperties {
 unsafe impl SharedCreepProperties for Creep {}
 unsafe impl SharedCreepProperties for PowerCreep {}
 
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub enum LineStyle {
+    Undefined,
```
daboross, line 268: I don't think this will serialize to the right value - it will make the string "undefined", rather than an actual JavaScript undefined value.

The only way I know of to actually serialize to `undefined`, though, is by leaving a value out entirely. I think we could do this if we added a `is_undefined()` method to `LineStyle`, and used [`skip_serializing_if`](https://serde.rs/field-attrs.html#skip_serializing_if) like so in `PolyStyle`:

```rust
#[serde(skip_serializing_if = "LineStyle::is_undefined")]
line_style: LineStyle
```

src/objects/creep_shared.rs
```diff
@@ -262,14 +262,79 @@ pub unsafe trait SharedCreepProperties: RoomObjectProperties {
...
+            fill: "#ffffff".to_string(),
+            opacity: 0.0,
+            stroke: "undefined".to_string(),
+            stroke_width: 0.1,
+            line_style: LineStyle::Undefined,
+        }
+    }
+}
+impl PolyStyle {
+    pub fn fill(mut self, fill: String) -> Self {
```
daboross, line 305: It'd be nice to have a short doc comment on each of these methods - I think we can just copy the descriptions from https://docs.screeps.com/api/#RoomVisual.poly.

src/objects/creep_shared.rs
```diff
@@ -338,15 +404,15 @@ where
     }
 
     /// Sets cost callback - default `|_, _| {}`.
-    pub fn cost_callback<'b, F2>(self, cost_callback: F2) -> MoveToOptions<'b, F2>
+    pub fn cost_callback<'b: 'a, F2>(self, cost_callback: F2) -> MoveToOptions<'a, F2>
```
daboross, line 407: Looks like there's a lifetime change included here that isn't used anywhere. Is this intentional, or could it be put back?

It looks like the kind of thing you'd need if you stored borrowed strings in PolyStyle - which is valid, but since we aren't doing it I don't _think_ it should be necessary.

Azaril:
Just a heads up I have the full set of RoomVisual types mapped over but not separated out.

For what it's worth, you don't need to specify the defaults on the styles. If the values are missing they get filled on the JS side during rendering if they're missing. We likely want to defer to that as it's cheaper and more correct.

drewcrawford:
@Azaril If a full solution on the RoomVisual stuff is a forthcoming PR, I would be in favor of tabling this one

Azaril:
> @Azaril If a full solution on the RoomVisual stuff is a forthcoming PR, I would be in favor of tabling this one

I've pushed a PR for the full API here: https://github.com/rustyscreeps/screeps-game-api/pull/306/

Azaril:
Closing this as the API came in with my PR.

Issue #279, opened 2020-02-01
RoomPosition::name() doesn't work in simulation mode


smessmer:
The correct room name for simulation mode would be "sim" while RoomPosition::name() returns "E0S0". This breaks several APIs, for example `creep.move_to(RoomPosition)` when used in simulation mode.

Tested in the JavaScript console where `creep.move_to(x, y, "sim")` works but `creep.move_to(x, y, "E0S0")` fails in the same way the rust API is failing.

PR #294, opened 2020-04-17
Improve the situation in the simulation mode

drewcrawford:
See #279

RoomName does not work in simulation mode, due to the way it is packed.  The room is named "sim", which does not unpack correctly into the integer.  There may be some clever way to reserve a value for it, but I am not sure how that would interface with screeps on the JS side.

I gather that sim mode is not used very heavily, although I find it useful for my problems.  Here are some hacks I've accumulated to make this work better

* new API to detect simulation mode in bots so they can special-case their own behavior
* game::room::get() will try `sim` if it can't find the room you asked for.  In sim mode, this is the only valid room, and in non-sim mode it should be `None`

### DIFF
--- a/src/local/room_name.rs
+++ b/src/local/room_name.rs
@@ -54,19 +54,22 @@ impl fmt::Display for RoomName {
     /// [`RoomName::new`]: struct.RoomName.html#method.new
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let x_coord = self.x_coord();
-
-        if x_coord >= 0 {
-            write!(f, "E{}", x_coord)?;
-        } else {
-            write!(f, "W{}", -x_coord - 1)?;
-        }
-
         let y_coord = self.y_coord();
 
-        if y_coord >= 0 {
-            write!(f, "S{}", y_coord)?;
+        if self.packed == 0 {
+            write!(f, "sim")?;
         } else {
-            write!(f, "N{}", -y_coord - 1)?;
+            if x_coord >= 0 {
+                write!(f, "E{}", x_coord)?;
+            } else {
+                write!(f, "W{}", -x_coord - 1)?;
+            }
+
+            if y_coord >= 0 {
+                write!(f, "S{}", y_coord)?;
+            } else {
+                write!(f, "N{}", -y_coord - 1)?;
+            }
         }
 
         Ok(())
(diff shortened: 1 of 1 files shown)


daboross:
What would you think of using some different sentinel value for `sim` here, rather than including a double check? I'm thinking since worlds are capped to around 100x100, we could have the `"sim"` be represented as `E127S127`, and do that mapping in parse and reproduction.

It wouldn't be much better than this solution, but it could at least ensure we don't need to perform extra checks if the room name is known not to be sim. I'm fairly hesitant to approve changes adding checks which will never be hit in production deployment.

I think it might be good to put this workaround, or maybe all of "sim" handling, behind a feature flag? That way we could have the code to handle it will without impacting builds destined to run on the server at all.

drewcrawford:
One thing I am unclear about is the cost of branching in this context.  Ordinarily a branch that is always taken, ought to be free with a competent branch predictor.  It's less clear to me what happens in screeps when you have bots constantly starting and stopping, does it blow out the predictor and you have to pay again?

Reserving a value in `RoomName` for `sim` would be cleaner, but it puts 2 checks in the path of using a `RoomName`, one when it's created and the other when it's destroyed.  Depending on the branch predictor, that's either twice as free or twice as expensive.

If it's twice as free, we don't need to gate it.  If it's twice as expensive we should probably gate a single-check solution, unless we don't care about the sim performance.

shanemadden:
For what it's worth the packed positions used by `sim` numerically map to positions in `W127N127`, it might work with a little less effort than `E127S127` if we go that route.

daboross:
> One thing I am unclear about is the cost of branching in this context. Ordinarily a branch that is always taken, ought to be free with a competent branch predictor. It's less clear to me what happens in screeps when you have bots constantly starting and stopping, does it blow out the predictor and you have to pay again?

That's a good point! I was thinking having a sentinel value would make the check somehow faster than a null check in JS, but that really doesn't make much sense when I think about it from that perspective. Thanks for catching my logic on that.

The only other advantage of a sentinel would be that it would work for positions too, and not just for getting rooms. Right now passing in a room name to create a new position and then using that position will, I believe, also fail miserably.

> For what it's worth the packed positions used by `sim` numerically map to positions in `W127N127`, it might work with a little less effort than `E127S127` if we go that route.

If the server uses W127N127 as its sentinel, we could do that, and I think get most of the position stuff working for free?

ASalvail (approved):

PR #295, opened 2020-04-17
Add max_cost to pathfinding options

drewcrawford:
`max_cost` is not being sent as an argument to pathfinder, so passing a value has no effect.  This improves performance on clients that set a non-default value.

Also, if you are me, you wrote code like this:

```rust
let mut best_result;
while condition {
    let result = screeps::pathfinder::search_many(&candidate.pos(), goals, options_with_max_cost);
    if !result.incomplete { 
        best_result = result;
    }
}
```

and then are puzzled when your search picks a ridiculous solution instead of the sensible one

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: rust
-sudo: false
 cache: cargo
 rust:
 - stable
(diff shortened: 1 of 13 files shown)


shanemadden:
@drewcrawford Would you mind merging the current master branch into your branch so that the build will succeed and we can merge this in? Thanks!

Azaril:
I merged the master branch and just waiting on CI checks.

PR #296, opened 2020-04-17
Make memory::get generic across error

### LINKED ISSUE #271: Make MemoryReference::get generic wrt Error
I was reminded by @internetimagery of a fix I made a while ago: 17d45721e44d9429fa13325f43056a5e5503ff3e
My memory is a bit hazy, since it was so long ago, and I forget why I held off on submitting it. If I remember correctly, it was to fix problems with implementing `Serialize` and `Deserialize`. If this is desired, should other parts also be changed in the same way, i.e. `get_path`?

drewcrawford:
resolve #271

### DIFF
--- a/src/memory.rs
+++ b/src/memory.rs
@@ -238,9 +238,9 @@ impl MemoryReference {
     /// info!("final position: {}", pos);
     /// creep.move_to(&pos);
     /// ```
-    pub fn get<T>(&self, key: &str) -> Result<Option<T>, ConversionError>
+    pub fn get<T, E>(&self, key: &str) -> Result<Option<T>, E>
     where
-        T: TryFrom<Value, Error = ConversionError>,
+        T: TryFrom<Value, Error = E>,
     {
         let val = js! {
             return (@{self.as_ref()})[@{key}];


src/memory.rs
```diff
@@ -238,9 +238,9 @@ impl MemoryReference {
     /// info!("final position: {}", pos);
     /// creep.move_to(&pos);
     /// ```
-    pub fn get<T>(&self, key: &str) -> Result<Option<T>, ConversionError>
+    pub fn get<T, E>(&self, key: &str) -> Result<Option<T>, E>
```
daboross, line 241: I think this could avoid the extra generic parameter if we just return `Result<Option<T>, <T as TryFrom<Value>>::Error>`. Since `Error` is an associated type, there's necessarily only one `Error` type per  `TryFrom<T>` implementation - so the `E` here would only ever have exactly one possible type.

daboross:
```suggestion
    pub fn get<T>(&self, key: &str) -> Result<Option<T>, <T as TryFrom<Value>>::Error>
```

I appreciate your work doing this! Only reason I suggest this is since I think it'd be a better solution. (but I really haven't been too involved in the project lately, thus I'm just commenting here rather than actually making the change)

src/memory.rs
```diff
@@ -238,9 +238,9 @@ impl MemoryReference {
...
     where
-        T: TryFrom<Value, Error = ConversionError>,
+        T: TryFrom<Value, Error = E>,
```
daboross, line 243:
```suggestion
        T: TryFrom<Value>,
```

PR #297, opened 2020-04-26
Add resource filter to market::get_all_orders

shanemadden:
The base game has an optimization for filtering by resource on `Game.market.getAllOrders` - it uses a 'full' lodash filter after retrieval but also grabs the `resourceType` key from that filter to limit what it's retrieving: https://github.com/screeps/engine/blob/d24939e8b5f44af3cefc0ca910a01e27e2aa2ae2/src/game/market.js#L37

This changes our `get_all_orders` interface to accept an `Option<MarketResourceType>` - sending `None` gets the normal behavior, while sending a specific resource will utilize that optimization, spending significantly less CPU time.

### DIFF
--- a/src/game/market.rs
+++ b/src/game/market.rs
@@ -246,17 +246,16 @@ pub fn extend_order(order_id: &str, add_amount: u32) -> ReturnCode {
 /// is available and will reduce the CPU cost compared to getting all orders
 pub fn get_all_orders(resource: Option<MarketResourceType>) -> Vec<Order> {
     match resource {
-        Some(resource_type) => match resource_type {
-            MarketResourceType::Resource(ty) => js_unwrap! {
+        Some(resource_type) => {
+            let resource_num = match resource_type {
+                MarketResourceType::Resource(ty) => ty as u32,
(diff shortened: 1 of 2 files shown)


src/game/market.rs
```diff
@@ -242,9 +242,24 @@ pub fn extend_order(order_id: &str, add_amount: u32) -> ReturnCode {
...
 /// Get all orders from the market
 ///
-/// Contrary to the JS version, filtering should be done afterwards.
-pub fn get_all_orders() -> Vec<Order> {
-    js_unwrap!(Game.market.getAllOrders())
+/// Full filtering support is not available, but filtering by resource type
+/// is available and will reduce the CPU cost compared to getting all orders
+pub fn get_all_orders(resource: Option<MarketResourceType>) -> Vec<Order> {
+    match resource {
+        Some(resource_type) => match resource_type {
```
daboross, line 249: I think we could avoid multiple js_unwraps in this case if we used `as u32` in each branch, and stored the result in a variable? Not sure if that would actually compile any differently, though.

PR #299, opened 2020-04-27
Allow for returning 'false' to pathfinder to prevent room pathing.

Azaril:
There are 3 return options for multi-room pathfinding functions:
1. Return a new cost-matrix. This value is used instead of the passed in one.
2. Return false which prevents the path finder from using this room.
3. Returning any other value results in the default cost-matrix being used. (This only contains room terrain I believe.)

This requires splitting all pathfinding related functions in to ones that either operate on a single room or deal with multiple room pathing. There is still a common set of options that are represented by FindOptions. To support both multi and single room, FindOptions is now generic with respect to the type returned by the cost callback. To limit this to sane values a tag of RoomCostResult is required.

### DIFF
--- a/src/local/room_position/game_methods.rs
+++ b/src/local/room_position/game_methods.rs
@@ -4,7 +4,7 @@ use crate::{
     game,
     local::RoomName,
     objects::{FindOptions, Flag, HasPosition, LookResult, Path},
-    pathfinder::CostMatrix,
+    pathfinder::{SingleRoomCostResult, CostMatrix},
 };
 
 use super::Position;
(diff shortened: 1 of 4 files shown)


Azaril:
Thanks for taking a look at the feedback.

Here's the code in the pathfinder that responds to 'false': https://github.com/screeps/driver/blob/e6849f9e2db9495e4344f3ac5d6dd3264e8fb3c6/lib/path-finder.js#L123

Room.findPath does call through the pathfinder but has an early out it the current room is not the same as the target room. It also has a different callback wrapper that doesn't handle a return value of 'false' which arguably is busted. This accounts for the difference between the pathfinder and the room APIs. See: https://github.com/screeps/engine/blob/92a3fffee65d547f12f5a99940c6155d9482a254/src/game/rooms.js#L248

In practicality the Room.findPath is odd in usage due to it early bailing if the rooms are different but however still allows the path being returned to traverse multiple rooms, see: https://github.com/screeps/engine/blob/5eec8a935db9715845ec5bdad1249fd9b8fe8e3d/src/game/rooms.js#L841 This needs an engine change and then we could unify the return results to multi-room everywhere.

The RoomPosition findPathTo delegates to the Room so in any case this API has to be identical to the room one.

Azaril:
I'm working through some more changes on this still to get it in to better shape.

daboross:
It could be nice to unify the APIs, but I think at this point that would be a pretty big breaking change for the engine, and I don't know if it would be worth it?


I think that Room.findPath and friends are useful in the beginning, but are almost always bypassed later on. Not being able to ignore a room is one problem, but even if we fixed that, they still trim all returned paths to one room, and always run the engine's cost calculation code once per room per tick. There's a lot of room for improvement there: path caching for multiple rooms is a big CPU improvement, as is caching calculated cost matrices over multiple ticks. In practice, at least when I was more active, I found most  players either used [Traveller](https://github.com/bonzaiferroni/Traveler) or built their own movement wrapper. And building a movement wrapper is a lot easier around PathFinder.findPath, just because of how it actually exposes all the information.

`Room.findPath` has a number of nice things going for it, including per-tick-cached calculated cost matrices, and automatic movement with moveTo, but almost every part of it is sub-optimal overall :/


Sorry to ramble! I think my main point is that there would probably be diminishing returns trying to get `Room.findPath` to be a good multi-room usable API. But if you want to take that on, feel free - this is just my experience.

Azaril:
So I think that's a totally reasonable perspective - this is a significant breaking change with questionable benefit for people that have existing code.

I'm working on essentially an equivalent to traveler but built in rust and I'm using some of these changes as building blocks. Let's put this PR on hold until there's something paired with it that brings some value to the community versus just API cleanup.

Azaril:
Closing this until there's more pressing need for it.

PR #304, opened 2020-04-29
Prevent panic when store_free_capacity returns a negative value

shanemadden:
Power creep `OPERATE_STORAGE` expiring can leave a storage with less capacity than resources, which causes `.store.getFreeCapacity()` to return a negative number, potentially causing a panic. Normalizes the return values to return 0 instead of panicking.


shanemadden:
Hrm.. potentially? I was thinking '0 is good enough, that's definitely the _"free capacity"_ when it's overflowing - that's not worth incurring conversion pain when people are comparing this plus a breaking change' but think in most cases where people are using it they're likely comparing to an independent threshold. There's.. actually only one spot in my code that would need to check whether this is negative then convert to a u32.  Could definitely go that way!

PR #305, opened 2020-04-30
Prevent panic in calls to mineral_type() when labs are empty

shanemadden:
Labs will return undefined for `mineralType` when empty - this changes our `mineral_type()` function to return an `Option<ResourceType>` to prevent panic when that's encountered.

### DIFF
--- a/src/objects/impls/structure_lab.rs
+++ b/src/objects/impls/structure_lab.rs
@@ -1,11 +1,25 @@
+use stdweb::Value;
+
 use crate::{
     constants::{ResourceType, ReturnCode},
     objects::{Creep, StructureLab},
+    traits::TryFrom,
 };
 
 impl StructureLab {
     pub fn mineral_type(&self) -> Option<ResourceType> {
-        js_unwrap!(__resource_type_str_to_num(@{self.as_ref()}.mineralType))
+        let mineral_v = js! {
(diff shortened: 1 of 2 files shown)


src/objects/impls/structure_lab.rs
```diff
@@ -1,11 +1,25 @@
...
+            const mineral = @{self.as_ref()}.mineralType;
+            if (mineral) {
+                return __resource_type_str_to_num(mineral);
+            }
+        };
+        match mineral_v {
+            Value::Number(_) => {
+                Some(ResourceType::try_from(mineral_v).expect("lab resource unknown."))
+            }
+            _ => None,
```
Azaril, line 21: Is it worth converting Value::Undefined to None and panic'ing on any other value?

ASalvail: Panics are bad. I'd suggest returning a `Result` if you want to announce an unexpected value.

shanemadden: I think `__resource_type_str_to_num` ought throw up an error that'll crash the tick on the js side when a value is present in `mineralType` that doesn't map to numbers the library recognizes, so we shouldn't get bad data at this stage generally; it might be clearer from a code perspective to be explicit about the mapping though?

ASalvail (approved):

PR #306, opened 2020-05-01
Room visual API.

Azaril:
This includes all the visual types and formatting options for the room visual API.

It directly calls the console addVisual to rather than more expensive calls through the RoomVisual API. This allows most of the operation to stay on the Rust side of the boundary until required.

For large batches of visuals it's actually faster to serialize the batch to a string, serialize that across and re-hydrate with JSON.parse. However, that's a tradeoff that can only really be made with knowledge of the data and likely not appropriate to put in the API layer itself. To enable this scenario there is the Visual enum that can be used for storage as well as appropriate serde annotations for all types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,9 @@ Unreleased
   since this is optimized in the server code (breaking)
 - Add `RoomTerrain::get_raw_buffer_to_array` to load a room's terrain into an existing `[u8; 2500]`
 - Change `MemoryReference::get` to return a generic error type
+- Add `RoomVisual`, rendering primitives (`Circle`, `Line`, `Rect`, `Poly`, `Text`).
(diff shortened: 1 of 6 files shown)


Azaril:
I had opted to only provide an implementation of 'default' instead of 'new' for the styles as there are default values that get filled in by the game API. The usage looks like this:

`CircleStyle::default().fill("green").opacity(1.0)`

I am not sure if that's the idiomatic Rust way of doing it so open to more feedback.

Azaril:
> The other thing I wasn't completely sure on is the positional args being used as floats - initially I was thinking 'I don't want to convert these from u32 to a float' but realized we probably do want that, to allow for more fine-grained positioning of shapes?

This is a bit of a pain to use but is part of the underlying API. I've typically created 'Into/From' implementations as needed to get in to the right format for my data types.

PR #307, opened 2020-05-02
Add GCL and GPL total point calculation functions

shanemadden:
The progress values returned by the game's GCL and GPL functions only shows number of points of progress toward the next level; if you're looking to calculate your total points, you also need to calculate the absolute lifetime point total for your current level.

It's a little iffy whether this should be in the library directly, but I figured I'd propose 'em after needing them for my own statistics - the formulas are not documented, so one needs to root around in the engine code or get the right answer in slack, then implement and test these functions, to be able to get at fairly basic 'should be in a constant' information.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,8 @@ Unreleased
 - Add `MoveToOptions::visualize_path_style`to allow for path visualization of movement system.
 - Change `HasStore::store_free_capacity` to return `i32`, handling potential negative values due
   to expiration of `OPERATE_STORAGE`
+- Add `game::gcl::total_for_level` and `game::gpl::total_for_level` which calculate the total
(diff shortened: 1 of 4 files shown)


shanemadden:
Updated with links, and realized there's internal constants to lean on for numbers here, so I switched it to do that.  In the process I changed `GCL_POW` to f64 from f32 - when I was using it as f32 the values were being calculated slightly less accurately, for example the number of points needed GCL 16->17:

    111357104.2997279 - calculated in rs with f32 GCL_POW
    111357070.7641989 - calculated in rs with f64 GCL_POW
    111357070.76419878 - reported by js from progressTotal

ASalvail:
Indeed, you can't represent accurately integers above 16'777'217 with `f32` (meaning that's where you start to see gaps between integers). Feel free to merge!

Issue #244, opened 2019-09-07
ObjectId encoding subtly wrong for LokIJS-backed private servers


daboross:
Turns out object IDs generated by LokIS have the possibility of starting with a `'0'`, and our current ObjectId implementation doesn't store this information. `"0df4aea318bd552"`, for instance, will be parsed fine, but we'll reproduce it as `"df4aea318bd552"` which the server won't recognize.

When doing research for ObjectId, I found a [`generateId` function that was part of the engine's _mock code_], and skipped over the [actual idGen function](https://github.com/screeps/storage/blob/2037ec3db916b947dd5df24986313ce518f9b913/lib/db.js#L106-L117) used.

Some ideas (from myself and mostly Dessix) for solving this:

- always zero-pad output to 15 digits
- store a length or number of 0 bits in the data
- config variable to enable padding, used when deploying to LokiJS servers
- if resolving a non-zero-padded ID fails, try with padding.
  
  if that succeeds, set a global variable and pad all IDs in the future

Storing a length would probably be the most reliable solution, but it also imposes an extra cost for storing length, and that cost will never be necessary on MongoDB-backed servers. Always zero-padding to 15 digits would work, but wouldn't be very future-proof for different ID generation strategies (or for 505 years in the future, when the id gen mechanism will rotate over and start producing 16-digit IDs).

A config variable is viable, but has the same problems as zero-padding IDs, and it introduces a UX cost - users would have to do something extra to deploy to private servers.

The last option seems a bit costly at first, but since it should only ever be used in private servers running LokiJS, maybe that cost is alright? I think most people who are either measuring CPU usage or wanting the fastest could should already be running MongoDB private servers.

ASalvail:
Zero-padding the string when it's sent over to JS seems like the easiest solution and adds the least overhead. I must say I am not swayed by the argument about anything that happens later than 5 years in the future 😂

daboross:
That would be the easiest solution! :p

I'm still a bit afraid of baking things like this, but I guess it's unlikely that they'll change the format, even for private servers. If we do the "always zero-pad to 15 digits" option, it would break 505 years in the future, but also if any private server ever switches to IDs with fewer than 15 digits, which I might be more worried about?

ASalvail:
We'll address this when it happens.

Adam Salvail

PR #311, opened 2020-05-05
Support object_ids that start with 0

### LINKED ISSUE #278: ObjectId deserialization doesn't work if id has leading zero
I'm storing the object ids of a `Source` into the `screep.memory` and load it later to try and get the `Source` object back. This works great if the object id starts with `[1-9a-f]`. But if the object id starts with a `0`, it breaks. It serializes it without the leading `0` and later, when trying to load it, it complains that it doesn't find an object with this id.

I've tried several mechanisms of serializing (including the common way of doing `id.to_string()` and then `str_id.parse()`, but also a serde way with `id.to_u128()` and then `serde_json`) but none of them worked, leading me to believe that the broken code is actually the one creating object id instances from and `u128`.

In case it matters, here's the serde code:

```
creep.memory().path_set("task.source_id", serde_json::to_string(&sources_id.to_u128())?);

[...]

let source_id = creep.memory()
    .path_string("task.source_id")?
    .unwrap()?;
let source_id: u128 = serde_json::from_str(&source_id)?;
let source_id: ObjectId<Source> = ObjectId::try_from(source_id)?;
let source = game::get_object_typed::<Source>(source_id)?
    .unwrap()?;
```

drewcrawford:
Resolve #278

This lets us parse and roundtrip ids that start with 0 correctly, both on Rust and JS.

To do this, we pad the packed value into a 12-byte (24-character) string, which is the expected length on [mongodb](https://docs.mongodb.com/manual/reference/method/ObjectId/).  This works for me in both sim and production.

I'm kind of unclear on whether there are other backends that may not like this format or if there is any reason to keep the shorter format around.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,3 +42,5 @@ stdweb-derive = "0.5"
 
 [features]
 check-all-casts = []
+#Shorten ObjectID representations to their minimum size.  Does not preserve leading 0s, see #311 for details
+short-ids = []
--- a/javascript/utils.js
+++ b/javascript/utils.js
@@ -352,6 +352,19 @@ function pos_from_packed(repr) {
     return pos;
 }
 
+function object_id_from_packed_16(slice) {
+    // reconstruct string in JS
+    let res = "";
+    for (var i = 0; i < slice.length; i++) {
+       if (i > 0) {
+           res += slice[i].toString(16).padStart(8, "0");
+       } else {
+           res += slice[i].toString(16);
+       }
+    }
+    return res.padStart(15, "0");
+}
+
 function object_id_from_packed(slice) {
     // reconstruct string in JS
     let res = "";
--- a/javascript/utils.js
+++ b/javascript/utils.js
@@ -356,11 +356,7 @@ function object_id_from_packed(slice) {
     // reconstruct string in JS
     let res = "";
     for (var i = 0; i < slice.length; i++) {
-       if (i > 0) {
            res += slice[i].toString(16).padStart(8, "0");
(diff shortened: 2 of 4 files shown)


shanemadden:
Unfortunately I think loki-backed private servers still use 15 byte ids - see #244. We'll probably need to store the length of the string to solve reliably, as discussed there.  Or we could declare loki private servers unsupported, but I suspect we should probably avoid going down that road given the number of complaints we already get for being incompatible with sim ;)

drewcrawford:
Would it work to gate 15-byte mode behind a feature?  I don't think we would need to support both in a single build, and it would avoid figuring out how to pack this in the type itself and the associated marshal/demarshal check

shanemadden:
Yeah, a feature flag for loki compatibility would probably do the job just fine without having to stress about performance penalties, and doing one for sim has been one of the ideas proposed for the issues there - I think that'd probably work.

shanemadden:
Hrmm.. one last big tricky bit on this is that the default map on private servers retains the 'random' 15 byte ids on room objects, so those objects break things.

drewcrawford:
I believe 15 byte mode should be supported with the `short-id` feature.

ASalvail:
From @shanemadden 
> Oh right - it's that if you have a mixed set of ids (like if you use the default private server map, which has short ids, but then use a mongo backend which uses long ids) then a feature flag doesn't cut it. I was thinking we'd maybe need to go with a more onerous fix of having a value in the id which indicated how many characters to pad to

Closing for now.

PR #314, opened 2020-05-07
Add implementation of findClosestBy for an array of objects.

misterwise:
Admittedly very new to both Rust and stdweb, however this implementation works.  Would love any feedback on making it more efficient.


misterwise:
Actually, upon further discussion it was explained how iterators could achieve this with likely greater efficiency.

ASalvail:
I'd be curious to hear what you replaced it with!

misterwise:
objs.iter().min_by_key(|obj| obj.pos().get_range_to(&creep.pos()))

The realization being that these objects have already been brought in from js so better to avoid calls out to js again.

PR #315, opened 2020-05-13
Improve parsing JS pathfinder results

drewcrawford:
In some exceptional cases, JS returns a path result with null for cost e.g.
`{path: [], cost: null, ...}`
For example, this happens in the case there are no walkable tiles around the start position.

For reasons that are opaque to me, this may get bridged into stdweb::Value
as Number of Infinity.  See also https://docs.rs/stdweb/0.4.20/src/stdweb/webcore/number.rs.html#1 which suggests to me there are sharp edges in their handling of some conversions involving the "weird" floats.

In any case, we have a u32 in this slot, so unless we break API we need to produce a u32 for this answer.  Due to the reproduction case and informed by "Infinity" I suggest placing u32::MAX here.

See also #309 for a similar issue involving unexpected path results. Informed by that I have substantially improved the failure handling around parsing this field.

### DIFF
--- a/src/pathfinder.rs
+++ b/src/pathfinder.rs
@@ -423,11 +423,33 @@ where
                 heuristicWeight: @{heuristic_weight}
             })
         };
+        //In some cases, like if we are searching from a position with no walkable neighbors,
+        //js comes back with `null` here.
+        //For reasons opaque to me, this is bridged into f64::POSITIVE_INFINITY rather than stdweb::Value::Null
+        let cost: u32 = match js! {return @{&res}.cost;} {
+            stdweb::Value::Number(n) => {
+                //see if the number fits into u32
+                if let Some(u) = n.try_into().ok() {
+                    u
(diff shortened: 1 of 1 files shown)


src/pathfinder.rs
```diff
@@ -423,11 +423,33 @@ where
...
+        //js comes back with `null` here.
+        //For reasons opaque to me, this is bridged into f64::POSITIVE_INFINITY rather than stdweb::Value::Null
+        let cost: u32 = match js! {return @{&res}.cost;} {
+            stdweb::Value::Number(n) => {
+                //see if the number fits into u32
+                if let Some(u) = n.try_into().ok() {
+                    u
+                } else if let Some(f) = TryInto::<f64>::try_into(n).ok() {
+                    if f.is_infinite() {
+                        std::u32::MAX
```
ASalvail, line 436: Any reasons not to switch to an `Option<u32>`? 

In any case, if we go this route we'd need to document this somewhere as someone might still try to follow that path if the cost is a valid one.

PR #317, opened 2020-05-27
Make .get_path generic

drewcrawford:
Similar to #302 but for a different API

### DIFF
--- a/src/memory.rs
+++ b/src/memory.rs
@@ -256,9 +256,9 @@ impl MemoryReference {
     /// `undefined`, and `Err` if incorrect type.
     ///
     /// Uses lodash in JavaScript to evaluate the path. See https://lodash.com/docs/#get.
-    pub fn get_path<T>(&self, path: &str) -> Result<Option<T>, ConversionError>
+    pub fn get_path<T>(&self, path: &str) -> Result<Option<T>, <T as TryFrom<Value>>::Error>
     where
-        T: TryFrom<Value, Error = ConversionError>,
+        T: TryFrom<Value>,
     {
         let val = js! {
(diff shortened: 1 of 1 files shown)


ASalvail:
@drewcrawford I remember having to deal with this some time ago. The `TryFrom<Value>::Error` is specified by `stdweb` but (memory is a bit hazy here) its visibility, from what I remember, wasn't meant to be public. That was why we made a homemade version of the `ConversionError`.

What do we get by changing it this way?

drewcrawford:
I think you are referring to this issue https://github.com/koute/stdweb/issues/280, which is a case where the implementation of the type conversion is statically known (and I believe it is also ourselves in that example, although I may be mistaken...)  

The underlying problem here involves when the conversion implementation is unknown, and the type is a usertype.   There is one case with repro steps in [this comment](https://github.com/rustyscreeps/screeps-game-api/issues/271#issuecomment-577932282) although that's `get` (#302), this is for `get_path`. Specific to this case, types that are `:JsSerialize` have a conversion to `Value` with error `stdweb::ecosystem::serde::ConversionError` but in this method signature we require error to be `stdweb::webcore::value::ConversionError` (the expansion of `ConversionError` in this context), so that will not pass typecheck.

More broadly, for the purposes of getting an arbitrary user type out of memory, it doesn't really matter whether the typeconversion implementation we use is serde's, stdweb's, or some homebrew `TryInto` with ~~`&'static str`~~ a fancy boutique error type.  Ultimately we just need to return whatever error encountered so the caller can ~~panic~~ perform appropriate handling.

ASalvail (approved):

PR #318, opened 2020-05-27
Add an IsStructure trait and related functionality

drewcrawford:
I have recently been writing construction planners.  Mostly, planning is specific to a single `structure_type`; for example a wall planner or a mine planner are intimately involved with the details of walls or mines respectively.  However, some functionality is similar across `structure_type`, like checking if a structure exists or is damaged.

Currently we have e.g. `StructureSpawn`, the specific type, and `Structure`, the general type.  For the purposes of boxing API results, this makes some sense, but there are a few issues for a planning usecase:

1.  Writing code broadly against `Structure` involves a lot of boxing/unboxing between specific and general types at runtime.  In my situation there are bajillions of these in a single logical planning operation which performs poorly.
2.  In practice, the planner wants to know about "approximately 1" `structure_type`, which is statically known, and this boxing introduces a lot of panic opportunities for logic errors that could be statically checked.

This PR introduces a trait `IsStructure` to which structures conform, allowing generic code to be written across `<S: IsStructure>` when the structure type is known, rather than across the runtimey `Structure`.  The `IsStructure` behavior is fairly minimal, is mostly what I needed for my case, and can be extended later.

Personally, I would prefer a more invasive change than this one – perhaps renaming `Structure` to `AnyStructure` and implementing a trait on it via [enum_dispatch](https://docs.rs/enum_dispatch/0.3.0/enum_dispatch/).  However, that would be much more invasive than this PR so I decided to start with something compatible.  Since JS doesn't fuss about e.g. "which" `.pos()` you call, I'm not sure how much strong typing matters internally, vs just exposing a typed interface ourselves.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -782,3 +782,27 @@ macro_rules! mem_set {
         compile_error!(concat!("Unexpected usage of mem_set! usage: ", stringify!($($not_valid)*)))
     }
 }
+
+macro_rules! impl_is_structure {
+    ($structure_struct_type: ty, $structure_type_pattern: pat,$c: ident,$structure_type: expr) => {
+        impl std::convert::TryFrom<Structure> for $structure_struct_type {
+            type Error = ConversionError;
+            fn try_from(s: Structure) -> Result<Self, Self::Error> {
+                match s {
+                    $structure_type_pattern => Ok($c),
+                    wrong => Err(ConversionError::Custom(format!(
+                        "Expected {}, got {:?}",
+                        stringify!($structure_struct_type),
+                        wrong.structure_type()
+                    ))),
+                }
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


src/macros.rs
```diff
@@ -782,3 +782,27 @@ macro_rules! mem_set {
         compile_error!(concat!("Unexpected usage of mem_set! usage: ", stringify!($($not_valid)*)))
     }
 }
+
+macro_rules! impl_is_structure {
+    ($structure_struct_type: ty, $structure_type_pattern: pat,$c: ident,$structure_type: expr) => {
+        impl std::convert::TryFrom<Structure> for $structure_struct_type {
```
ASalvail, line 788: `TryFrom` was stabilized?

PR #319, opened 2020-05-28
Intershard memory edge case fixes

shanemadden:
`InterShardMemory` - heavy sigh.

- On private servers, it doesn't exist at all.  Handle the cases where it's undefined by returning `None`
- On MMO, when you haven't set data for a segment it's undefined and can cause a panic; handle that with an `Option` wrapper

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,10 @@
 Unreleased
 ==========
 
-- Change `game::inter_shard_memory` functions to avoid panicking on private servers where the interface doesn't
-  exist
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 Unreleased
 ==========
 
+- Change `game::inter_shard_memory` functions to avoid panicking on private servers where the interface doesn't
+  exist
```
ASalvail, line 5: Those are breaking changes.

PR #321, opened 2020-06-06
Update constants doc comments for numbers, add controller_structures, and other related additions

shanemadden:
- Added doc comments to most of the stuff in `numbers.rs` to point to how they're used.  Some of these got a little verbose, but the formulas should be stable, and having a reference for stuff like the tower falloff formula which is otherwise kinda a pain to track down in code or patch notes seems valuable
 - Added `controller_structures` function to `StructureType` enum, which handles translation of `CONTROLLER_STRUCTURES` constant by returning how many of that structure can be constructed at a given controller level
 - Added `cancel_delete` to `AccountPowerCreep` to call delete with the optional cancel parameter to enable stopping deletion of power creeps
 - Added `MARKET_MAX_ORDERS`, `MARKET_ORDER_LIFE_TIME`, and `HARVEST_DEPOSIT_POWER` constants
 - Corrected `stronghold_rampart_hits` for an upstream rebalancing and `POWER_SPAWN_HITS` / `RAMPART_HITS_MAX_RCL5` for typos
 - Removed `INVADER_CORE_EXPAND_TIME` which had already been replaced by per-level `invader_core_expand_time`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,18 @@
 Unreleased
 ==========
 
+- Add `AccountPowerCreep::cancel_delete`, which allows cancelling pending deletion of a power
+  creep
+- Add `StructureType::controller_structures`, which translates the `CONTROLLER_STRUCTURES`
+  constant, the number of each structure allowed at a given RCL
+- Update `constants::stronghold_rampart_hits` function with updated values from rebalancing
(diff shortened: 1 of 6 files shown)


Azaril:
@shanemadden can you merge this to latest now CI is unblocked?

PR #322, opened 2020-06-06
Screeps decoration update new features and deprecations

shanemadden:
New features for Screeps 4.2 currently on PTR, not ready for merge but able to be used for anyone testing on the PTR.

Currently just the 3 new intershard resources and the new function and constant for pixel generation - I expect there will be more changes to come.


shanemadden:
Updated and works on the released MMO version, but the devs won't be adding stub functions for the cpu unlock and pixel interfaces, so we'll want those to fail in safer ways than currently implemented.

javascript/utils.js
```diff
@@ -207,6 +207,9 @@ function __resource_type_num_to_str(num) {
         case 83: return RESOURCE_EMANATION;
         case 84: return RESOURCE_ESSENCE;
         case 1001: return SUBSCRIPTION_TOKEN;
+        case 1002: return CPU_UNLOCK;
```
ASalvail, line 210: Is it just me or are all those constants gone from the screeps doc? We could just now encode this in an enum instead.

Not in this PR, but still.

src/game/map.rs
```diff
@@ -164,13 +163,12 @@ pub fn find_route_with_callback(
 
     // Type erased and boxed callback: no longer a type specific to the closure
     // passed in, now unified as &Fn
-    let callback_type_erased: &mut (dyn FnMut(RoomName, RoomName) -> f64) =
-        &mut callback_boxed;
+    let callback_type_erased: &mut (dyn FnMut(RoomName, RoomName) -> f64) = &mut callback_boxed;
```
ASalvail, line 166: Do you have a special `rustfmt` config? Line length should be pretty standard; I don't get why it's jumping around.

src/game/market.rs
```diff
@@ -1,303 +1,304 @@
...
+/// Provide a resource type to get history for using `Some(ResourceType)`, or
+/// get data for all resources by passing `None`
+pub fn get_history(resource: Option<MarketResourceType>) -> Vec<OrderHistoryRecord> {
+    match resource {
+        Some(resource_type) => {
+            match resource_type {
+                // workaround: Game.market.getHistory returns `{}` instead of `[]` when querying a resource type
+                // that has no history records
+                // Verify records are present otherwise return an empty array to prevent panics
+                // Workaround no longer needed if this PR gets merged: https://github.com/screeps/engine/pull/131
```
ASalvail, line 276: Can we track this in an issue instead of in the code?

ASalvail (approved):

javascript/utils.js
shanemadden: Hrmm not sure if I'm following - should still be in their docs? https://github.com/screeps/common/blob/master/lib/constants.js#L379

src/game/map.rs
shanemadden: Nothing custom, I'm just gonna keep blaming nightly builds ;)

javascript/utils.js
ASalvail: You're right, I checked their history and I imagined things. I thought the integers were related to similar constants in the js implementation. It seems we only use them to make the translation easier.

PR #323, opened 2020-06-13
Add the parse_display::Display trait to enums already using FromStr

shanemadden:
@Azaril noticed that while we can go from the js-constant-string representation of a resource type to a native enum without having to call the js helper functions, the reverse wasn't true - as far as I can tell we don't have a way to go back to these values (say, for stats recording or writing public segment data for other bots to read) without a round trip to js.  Adding the `parse_display::Display` trait to those enums which already have the `parse_display::FromStr` trait gives us that capability.


shanemadden:
Yup, all good there -

    info!("{}, {}", ResourceType::Energy.to_string(), serde_json::to_string(&ResourceType::Energy).unwrap());

results in `energy, 1` 👍

Azaril:
@shanemadden can you merge this to latest now CI is fixed?

PR #326, opened 2020-06-29
Pathfinding v2

Azaril:
This is another pass at adding back in full cost matrix return values.

- Adds the correct return values multi-room pathfinding that allows to make a room as impassable.
- Removes complexity for pathfinding callbacks that were being stored in thread local storage that can remain on the stack.
- Fixes callback leaks on the JS side caused by not dropping the callbacks.
- I'm proposing we take the breaking change as it seems like people are already using these changes as they're needed to make multi-room pathfinding work out of the box.

### DIFF
--- a/src/game/map.rs
+++ b/src/game/map.rs
@@ -5,7 +5,6 @@ use std::{borrow::Cow, collections, mem, str::FromStr};
 
 use num_traits::FromPrimitive;
 use parse_display::FromStr;
-use scoped_tls::scoped_thread_local;
 use serde::{
     de::{Deserializer, Error as _, Unexpected},
     Deserialize,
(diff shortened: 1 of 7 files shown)


src/game/map.rs
```diff
@@ -106,80 +105,84 @@ pub fn find_exit(from_room: RoomName, to_room: RoomName) -> Result<ExitDirection
...
+    };
+
+    // Type erased and boxed callback: no longer a type specific to the closure
+    // passed in, now unified as &Fn
+    let callback_type_erased: &mut (dyn FnMut(RoomName, RoomName) -> f64) =
+        &mut callback_boxed;
+
+    // Overwrite lifetime of reference so it can be passed to javascript. 
+    // It's now pretending to be static data. This should be entirely safe
+    // because we're only sticking it in scoped storage and we control the
```
daboross, line 172: Might be worth removing the reference to scoped storage?

I think the point still stands, it's just that we're trusting `Game.map.findRoute` to never persist our closure, rather than trusting `scoped_tls`.

src/pathfinder.rs
```diff
@@ -415,29 +500,47 @@ where
...
+    };
+
+    // Type erased and boxed callback: no longer a type specific to the closure
+    // passed in, now unified as &Fn
+    let callback_type_erased: &mut (dyn FnMut(RoomName) -> Value + 'a) =
+        &mut callback_boxed;
+
+    // Overwrite lifetime of reference so it can be passed to javascript. 
+    // It's now pretending to be static data. This should be entirely safe
+    // because we're only sticking it in scoped storage and we control the
```
daboross, line 516: Same here (same comment).

Issue #320, opened 2020-05-28
Event types EVENT_POWER and EVENT_TRANSFER not supported


shanemadden:
The `EVENT_TRANSFER` and `EVENT_POWER` event types have been added at some point since we implemented event support; the presence of these events in the log currently causes a panic.

`EVENT_TRANSFER` has a documented format, but I can't find any documentation for `EVENT_POWER` - will need to observe it to get the format right for a new struct.

shanemadden:
Fixed by #327

PR #327, opened 2020-08-25
Fix deserializing for Power and Transfer event.

wenjoseph:
`Room.get_event_log()` panics when the event is either `EVENT_TRANSFER` or `EVENT_POWER`.

This seems to fix the problem for me. Please take a look. Thanks,


shanemadden:
Thanks a bunch, this fixes #320 - I noticed it wasn't implemented but thought we needed to do all the struct and enum work as well!

shanemadden (approved):

PR #330, opened 2020-10-05
fix Position::get_direction_to

fuchsnj:
`Position::get_direction_to` always returned the opposite of the correct direction. This fixes it.


ASalvail (approved):
With a test even!

Thanks!

ASalvail:
Could you add a line to the `changelog.md` for the bugfix?

PR #332, opened 2020-11-19
Move changelog lines to proper section

alelazar:
My first rust PR! With a lot less rust than I'd like but eh. 

Mis-log is from https://github.com/rustyscreeps/screeps-game-api/commit/34cf6b35393c5d912ccaa1a00f4b1d59641eb0f9#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR58 looks like it was written before 0.8 but then merged after.


shanemadden (approved):
Thanks for cleaning up my mess ;)

PR #333, opened 2020-11-20
Convert from stdweb to wasm-bindgen

### LINKED ISSUE #156: Design of functions taking optional names
The main four here are `Room.createConstructionSite`, `Room.createFlag`, `RoomPosition.createConstructionSite` and `RoomPosition.createFlag`.

Currently, neither `createFlag` can be called without a name - but this is something that the API should allow for. `createConstructionSite` is translated as two functions: `create_construction_site` and `create_named_construction_site`.

We should decide what pattern we want all of these to follow.

Possibilities I can think of:

- one function taking `Option<&str>`
- one function taking `impl Into<Option<&str>>`
- two functions - one `create_x`, one `create_named_x`
- two functions - one `create_unnamed_x`, one `create_named_x`

### LINKED ISSUE #246: creep.id() panics if the creep is still spawning

### LINKED ISSUE #287: nodejs cargo test failing under 1.44 nightlies
Builds have been failing for several days due to a test failure introduced with the rust 1.44 nightly builds - the `cargo web test --verbose --nodejs` step fails with `thread 'main' panicked at 'internal error: no .js file found', .cargo/registry/src/github.com-1ecc6299db9ec823/cargo-web-0.6.26/src/cmd_test.rs:58:9` - seems like the js files are not being included in the build result artifacts, but I'm unsure if this is an issue in stdweb or rust itself - it doesn't seem likely to be our fault.

I'm planning on putting a PR in to comment out that nodejs test for now, while leaving this issue open to track re-enabling that test when it's not failing in the nightly versions anymore.

### LINKED ISSUE #308: Some objects have no id
Certain objects do not have an id, and this seems like a footgun.  For example, 

1.  Spawn a new creep
2.  During the same tick as 1, do `screeps::game::creeps::values().iter().map(|p| p.id())`
3.  Traceback

```
1:18:24 PM]caught exception: TypeError: Cannot read property 'length' of undefined
[1:18:24 PM]stack trace: TypeError: Cannot read property 'length' of undefined
    at object_id_to_packed (eval at exports.evalCode (blob:https://screeps.com/29a5a139-d638-4e92-a50d-359e08cec132:2:11757), :870:12)
    at eval (eval at exports.evalCode (blob:https://screeps.com/29a5a139-d638-4e92-a50d-359e08cec132:2:11757), :1101:107)
    at __cargo_web_snippet_fa761c1402d330c48e501b99c4185c8157f04119 (eval at exports.evalCode (blob:https://screeps.com/29a5a139-d638-4e92-a50d-359e08cec132:2:11757), :1101:138)
    at _ZN7screeps7objects5HasId2id17hd202e9200e20e409E (wasm-function[550]:0x2ee78)
```
Although this traceback is opaque, ultimately the root cause is a creep has no id during its first tick (of being spawned) and so screeps-game-api crashes trying to get it.  

It's not immediately clear how to fix this.  `HasId` has `fn id(&self) -> ObjectId<Self>` but when there is no id I don't think this can be implemented correctly, instead we need a signature like `Option<ObjectId<Self>>`.

If this is specific to `Creep` we could potentially introduce a new trait like `MaybeHasId`, although this would complicate generic interfaces.  However my suspicion is that it actually happens for most types under the right circumstances, and so we maybe ought to change `HasId` itself.  However I imagine that this is a nontrivial API breakage.

### LINKED ISSUE #316: tickLimit returns `Infinity` on sim
Calling `sim.tickLimit` used to work, but it no longer works after [this commit](https://github.com/rustyscreeps/screeps-game-api/commit/a7b28339dd99275ea6aa5712488243d40628e08f).

This is because `tickLimit` is actually `Infinity`, which cannot be bridged as an int.

### LINKED ISSUE #328: Add Map Visual Support
The map visual feature is live - [announcement post](https://screeps.com/forum/topic/3064/map-visuals) / [documentation](https://docs.screeps.com/api/#Game-map-visual), we should add support for it. It's similar to RoomVisual so some of that may be portable between both, but has some minor implementation differences.

### LINKED ISSUE #329: Explore use of wasm-bindgen instead of stdweb
The `wasm-bindgen` toolchain is much better supported than `stdweb` and is the "official" way to call into JS from Rust. As seen in https://github.com/rustyscreeps/screeps-game-api/pull/288, `stdweb` and `cargo-web` don't see a lot of attention. The `gerandom`/`rand` crates also had to disable testing for `cargo-web` due to lack of upstream support.

The issue https://github.com/screeps/screeps/issues/120 and [this feature request](https://screeps.com/forum/topic/3067/allow-access-to-node-s-textdecoder) descirbe the necessary first steps to supporting `wasm-bindgen`.

### LINKED ISSUE #331: Fatal runtime errors for all code built by Rust 1.48 and above
Rust nightly builds have been bumped up to 1.49, and when built with this version both my live bot's code and our sample begin throwing runtime stdweb conversion errors:

    RuntimeError: unreachable
        at rust_panic (wasm-function[178]:1)
        at _ZN3std9panicking20rust_panic_with_hook17h1347218c91d128f7E (wasm-function[170]:117)
        at _ZN3std9panicking19begin_panic_handler28_$u7b$$u7b$closure$u7d$$u7d$17hfae94c93fb3d0300E (wasm-function[181]:1)
        at _ZN3std10sys_common9backtrace26__rust_end_short_backtrace17hf812f49b8d150ee0E (wasm-function[180]:6)
        at rust_begin_unwind (wasm-function[27]:52)
        at _ZN4core9panicking9panic_fmt17h6aa2a8f84484b5f7E (wasm-function[24]:48)
        at _ZN4core9panicking5panic17hc886a4cb4479b06eE (wasm-function[14]:67)
        at _ZN148_$LT$stdweb..webcore..serialization..SerializedValue$u20$as$u20$core..convert..From$LT$stdweb..webcore..serialization..SerializedUntaggedF64$GT$$GT$4from17h1f45361611ee357cE (wasm-function[72]:14)
        at _ZN50_$LT$T$u20$as$u20$core..convert..Into$LT$U$GT$$GT$4into17h099fb7234c614c67E (wasm-function[71]:1)

If we're unlucky, this breakage will carry down to stable like it did with #287 👎 ; probably #329 will be our way forward since stdweb isn't showing any signs of life.

shanemadden:
Resolves #329.  🎊
Resolves #156, resolves #246, resolves #287, resolves #308, resolves #316, resolves #328, resolves #331

See the temporary readme file added in the root for how to get started using this branch; there's some manual steps to hack the wasm-pack output for now.

Remaining to-do items:

- [x]  Return codes - negative values aren't natively supported by bindgen for enums, so we're taking all return code returning functions as i8 right now. Implement better conversion for `ReturnCode` and change i8 function returns.
- [x]  Logging - currently just referencing web-sys's console functions, get a good example going with fern.
- [x]  Market - a bunch of return types need struct conversions instead of raw `Object`, and MarketResourceType conversion from JsString
- [x]  Memory root - need to figure out how to get bindgen to return this object directly from the global scope fresh on each access, which is.. surprisingly confounding to figure out how to do?  Document using the thing, and its children from creep/room/etc getters, using Reflect.
- [x]  portal destinations - return type needs struct conversion
- [x]  Room event logs - return types need struct conversion
- [x]  Typed find/look_for calls
- [ ]  final - For all object prototype functions where it's safe, add the bindgen `final` keyword to link the functions more efficiently
- [ ] structural - opposite of final, mark everything on the Game object, which gets re-created every tick, as explicitly structural even though that's currently the default
- [x]  RoomVisual
- [x]  MapVisual
- [x]  Builds - let's move to GH actions as we move to new wasm-pack testing toolchain, since travis has been problematic.
- [x]  Typed Object IDs - currently handling IDs as JsString without a local representation yet, work out how to deal with storing an object type along with an Id and resolving properly, preferably with support for at least both 24-char and 15-char 0-padded IDs (add to tests)
- [x]  Structure typing - we have an enum on structureType that can be used to get the object type as an enum, but not yet as way to get a `Structure` as the narrowly typed structure within it
- [x]  Chaining - types that support chaining in js like visual inputs should support chained calls
- [x]  Code deployment - update cargo-screeps to support build/deploy of bindgen codebases ~~(the hacks currently enshrined in the temporary readme)~~ https://github.com/rustyscreeps/cargo-screeps/pull/24
- [ ]  More doc examples - especially of 'tricky' stuff like passing Uint8Array::view() into a CostMatrix unsafely
- [ ]  Anything else I added a `todo` comment on and forgot about

Maybe:

- [x]  Trait implementations - bindgen has _opinions_ about how it hooks into the js inheritance hierarchy, does it make sense to re-implement our traits against the child object types or just expect passing a parent type AsRef as the API is now?
- [ ]  Terrain local representation, similar to the CostMatrix one, providing a fast rust-native interface to a `Vec<u8>` terrain buffer from js (notably with x/y flipped vs costmatrix pos_as_idx)


Cargo.toml
```diff
@@ -29,15 +28,13 @@ travis-ci = { repository = "rustyscreeps/screeps-game-api" }
...
 num-derive = "0.2"
 num-traits = "0.2"
-parse-display = "0.1"
-serde = {version = "1", features = ["derive"]}
+serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 serde_repr = "0.1"
-stdweb = "0.4"
-stdweb-derive = "0.5"
+wasm-bindgen = { version = "0.2", features = ["serde-serialize"] }
```
alelazar, line 37: yes

traxys:
I made a script to automate the manual steps for deploying the js files: https://github.com/traxys/screeps_base/blob/master/copy.sh
This builds on the wasm-pack-template adding code to replace the parts in the js file. It's not the prettiest thing but it works

shanemadden:
@traxys Thanks, that'll be really useful to save me a bunch of copy-pasting until we get cargo-screeps updated to do this stuff! 👍

juzi5201314:
I tried wasm-bindgen in `sim`, screeps doesn't seem to work with `Reflect`, which makes `JsHashMap` etc. not work.

There is no method in Reflect:
`error: Reflect.get is not a function.`

Dessix:
@shanemadden I'm looking to continue this work; the last commits were more recent than the checklist's last update- are these still the accurate state of affairs?

ASalvail:
I missed this! What an incredible milestone!
I might be tempted to give screeps another try now.

PR #336, opened 2020-12-18
Update PIXEL_CPU_COST for balance change

shanemadden:
Pixels [got more expensive](https://screeps.com/forum/topic/3099/game-cpu-generatepixel-change/109) - update pixel cost constant to 10,000 to reflect that.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -1,6 +1,6 @@
 //! Constants, most copied from [the game constants].
 //!
-//! Last updated on 2020-06-06, `3bc7337807d380ffb29b9dcec40fc19e707c2476` from
+//! Last updated on 2020-12-13, `e4589666113334bb1f967b9a5540b642141b6dab` from
 //! <https://github.com/screeps/common/commits/master/lib/constants.js>.
 //!
 //! Currently missing:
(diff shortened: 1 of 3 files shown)


src/constants.rs
```diff
@@ -1,6 +1,6 @@
 //! Constants, most copied from [the game constants].
 //!
-//! Last updated on 2020-06-06, `3bc7337807d380ffb29b9dcec40fc19e707c2476` from
```
birktj, line 3: The old commit id here is is quite old (https://github.com/screeps/common/commit/3bc7337807d380ffb29b9dcec40fc19e707c2476 was back in 2019) while the timestamp is newer, however it seems that all changes done since that commit have made their way into this library.

birktj (approved):
The new constant seems match the one in constants.js. All changes since the last noted commit have been included in the library. Changelog has been updated.

birktj:
For future discussion I wonder whether using `lazy_static!` to fetch constants from js would be an idea? This would of course have the downside of changing the type of constants from `const T` to `static Deref<T>` which would be more clunky to work with and not usable in `const` computations. However it would have the advantage of being able to adjust to changes in constants between various servers etc. Maybe for certain constants that have a high probability of changing?

ASalvail (approved):

shanemadden:
@birktj Yeah, we've kicked around some ideas in the past for dynamic constant loading and/or parsing and it's a feature we'd really like to have (especially as seasonal will continue to drive updates of the not-even-in-that-file seasonal constant sets) - maybe something at compile time, maybe with macros or in postprocessing in cargo-screeps - but the detriments of various methods have meant that we've never gotten anything implemented.

PR #342, opened 2021-03-23
Enable RoomVisual API on bindgen.

Azaril:


shanemadden:
Looks good and seems to be working, thanks again!

At some point we should look into potentially using the new import/export functions for performance, but their format seems to be newline-separated JSON objects instead of a single valid one, which makes it trickier to work with.

shanemadden (approved):

PR #345, opened 2021-04-11
Traits and feature improvements

Azaril:
- Remove duplicated code for structure types - use traits with blanket implementations where possible.
- Add typed 'find' constants and fix minimum set of functions.
- Add back in some old 'Structure' enumeration implementations like 'as_has_store'.
- Add 'maybe' traits for 'pos' and 'id' so that common cases (just getting pos) are much simpler in 99% cases.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -66,8 +66,9 @@ mod recipes;
 pub mod seasonal;
 mod small_enums;
 mod types;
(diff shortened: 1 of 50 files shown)


Azaril:
Closing as cleanup and more changes are in a newer pull request.

PR #346, opened 2021-04-14
Converting Position into using u8 for in-room x and y

khoover:


src/local/room_position/pair_utils.rs
```diff
@@ -5,28 +5,30 @@ impl Position {
...
 // implementation is not allowed as it'd be implementing it on a nested type,
 // and both the outer type (`(T, T)`), and the inner ones (`u8`) are from an
 // external crate.
+// TODO: We can do stuff like impl From<LocalCostMatrix> for Vec<u8>, so maybe
+// we can do the same here for more idiomatic code?
 
 impl Into<(u8, u8)> for Position {
     /// Turns this positions's in-room coordinates into a pair of `u8` values.
     #[inline]
     fn into(self) -> (u8, u8) {
```
Azaril, line 30: Can you call self.coords()? Is it worth having an 'into' for i8 as well?

khoover: I'm actually somewhat against having Into<(u8, u8)> and related at all, since it's ambiguous whether it's meant to be world coordinates or in-room. But it could use self.coords() here, just more explicit that this is the in-room when written like this.

Azaril: Agreed on the ambiguity so that's fine to ignore. Using coords just saves some type casting and is a bit cleaner but it's a minor nitpick.

PR #347, opened 2021-04-14
LocalCostMatrix improvements

khoover:
Some improvements to LocalCostMatrix that should increase performance, make it more usable, and cover a missing bounds check.

Making `bits` a [u8; 2500] does mean that a LocalCostMatrix has a size of 2.5kB instead of whatever small size the Vec has, so users may want to use a Box<LocalCostMatrix> wherever a LocalCostMatrix was used before.

### DIFF
--- a/src/local/cost_matrix.rs
+++ b/src/local/cost_matrix.rs
@@ -1,5 +1,7 @@
-use crate::objects::CostMatrix;
 use std::convert::TryInto;
+use std::ops::{Index, IndexMut};
+
+use crate::objects::CostMatrix;
 
 #[derive(Clone, Debug)]
 pub struct LocalCostMatrix {
@@ -114,6 +116,24 @@ impl From<CostMatrix> for LocalCostMatrix {
         LocalCostMatrix {
             bits: array.to_vec().try_into().expect("JS CostMatrix was not length 2500."),
         }
+    }
+}
+
+impl Index<(u8, u8)> for LocalCostMatrix {
+    type Output = u8;
+
+    fn index(&self, idx: (u8, u8)) -> &Self::Output {
+        assert!(idx.0 < 50, "out of bounds x: {}", idx.0);
+        assert!(idx.1 < 50, "out of bounds y: {}", idx.1);
+        &self.bits[pos_as_idx(idx.0, idx.1)]
+    }
+}
+
+impl IndexMut<(u8, u8)> for LocalCostMatrix {
+    fn index_mut(&mut self, idx: (u8, u8)) -> &mut Self::Output {
+        assert!(idx.0 < 50, "out of bounds x: {}", idx.0);
+        assert!(idx.1 < 50, "out of bounds y: {}", idx.1);
+        &mut self.bits[pos_as_idx(idx.0, idx.1)]
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/local/cost_matrix.rs
```diff
@@ -21,21 +25,39 @@ impl LocalCostMatrix {
...
     pub fn get(&self, x: u8, y: u8) -> u8 {
-        self.bits[pos_as_idx(x, y)]
+        self[(x, y)]
+    }
+
+    // # Safety
+    // Calling this method with x >= 50 or y >= 50 is undefined behaviour.
+    #[inline]
+    pub unsafe fn get_unchecked(&self, x: u8, y: u8) -> u8 {
+        debug_assert!(x < 50, "out of bounds x: {}", x);
```
Azaril, line 46: Might be simpler to add a function to get the reference to a entry in the cost matrix. That way you can move the checks and the unchecked get in to a single location versus duplicating.

pub unsafe fn get_entry(&mut self, x: u8, y: u8) -> &mut u8 {
    debug_assert!(x < 50, "out of bounds x: {}", x);
    debug_assert!(y < 50, "out of bounds y: {}", y);
    self.bits.get_unchecked_mut(pos_as_idx(x, y));
}

Azaril (commented):
I believe that LocalCostMatrix::new was already allocating the vector anyway so not sure there's any meaningful memory cost size change.

I like the Position based index as well - that's a much nicer API to deal with.

khoover:
> I believe that LocalCostMatrix::new was already allocating the vector anyway so not sure there's any meaningful memory cost size change.

It's more where the memory alloc happened; before, LocalCostMatrix::new was a small stack alloc and a 2.5kB heap. Now it's a 2.5kB stack.

src/local/cost_matrix.rs
khoover: I was thinking about having a `get_unchecked_mut`, but even then you'd need both due to `&self` vs `&mut self`.

Azaril: You're right - you'd still need both. It would still reduce the number of places the asserts are needed from 4 down to too so still worth it.

khoover: I'm still not sure how we'd get down to 2; there's the non-debug asserts in the indexing impls, which are there for correctness, and there's the debug asserts in the unchecked methods, which are there as testing aids. Can't replace one sort with the other, since they're for different purposes.

Azaril: I had missed the nuance on the asset vs debug_assert - you're using the debug_asset to reduce the cost in places where it shouldn't panic. The raw u8 indexing could come from user values and as such needs to be untrusted and panic.

The only way to factor it all out so there's 2 places using get_unchecked(_mut) would end up with the assert functions calling to ones that have debug_asset which is a bit gross. You could still combine the position indexing with the get/set functions.

Azaril (approved):

src/local/cost_matrix.rs
khoover: You mean have the indexing functions call in to the self.get_unchecked? That would work if it returned a reference, which I suppose it could do. Would be weird having get returning u8 and get_unchecked returning &u8 (similar with set vs get_unchecked_mut).

khoover:
I can't merge it.

PR #349, opened 2021-04-18
Bringing parity back for pathfinder, movement and a number of find APIs.

Azaril:
- Cleanup duplicated traits across structures.
- Move to using serde-wasm-bindgen instead of built-in bindgen serialization that uses stringification + JSON parsing.
- Implement pathfinder functions back to parity including options callbacks.
- Implement typed find methods. (Still missing 'Terrain' find.)
- Add strongly typed wrapper containers for 'Array' and 'HashMap' JS objects.
- Add 'as_has_store' back to structure enum type. (Other traits need adding still!)
- Add missing structure properties.
- Remove old stdweb code for functionality that is now implemented.

NOTE: This is not very well tested as there's a lot of surface area missing just to get back to parity. Once parity for critical functionality is done it will be simpler to test with existing bots.


src/local/room_name.rs
```diff
@@ -153,6 +155,55 @@ impl RoomName {
...
+
+impl Into<JsString> for &RoomName {
+    fn into(self) -> JsString {
+        let val: JsValue = self.into();
+
+        val.unchecked_into()
+    }
+}
+
+impl From<JsValue> for RoomName {
```
khoover, line 190: This and below should really be `TryFrom`, seeing as they could very easily fail.

Azaril: I'll move them to be an 'unchecked_into' since they're only intended to be used on the marshalling boundary and visible to this crate.

Azaril: I made them try_from but added 'expect' unwrapping in places it should never fail.

src/constants/find.rs
```diff
@@ -0,0 +1,520 @@
...
+    pub struct EXIT_RIGHT = (Find::ExitRight, RoomPosition);
+    pub struct EXIT_BOTTOM = (Find::ExitBottom, RoomPosition);
+    pub struct EXIT_LEFT = (Find::ExitLeft, RoomPosition);
+    pub struct EXIT = (Find::Exit, RoomPosition);
+}
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen]
+    pub type JsFindOptions;
```
shanemadden, line 268: `JsFindOptions` and `FindOptions` might fit better to implement in room.rs since they're specific to the functions on that object type (their old implementations are still in there commented out at the moment).

shanemadden (approved):

PR #350, opened 2021-04-18
RoomCoordinate and RoomXY wrapper types

khoover:
This is meant to be Position for room-agnostic coordinates (e.g. cost-matrix cells). Puts all of the bounds checking into a single place.

### DIFF
--- a/src/local/cost_matrix.rs
+++ b/src/local/cost_matrix.rs
@@ -3,23 +3,16 @@ use std::convert::TryInto;
 use std::iter::IntoIterator;
 use std::ops::{Index, IndexMut};
 
+use serde::{Serialize, Deserialize};
+
 use crate::objects::CostMatrix;
 
-use super::Position;
+use super::{Position, xy_to_linear_index, linear_index_to_xy, RoomCoordinate, RoomXY, ROOM_AREA};
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct LocalCostMatrix {
-    bits: [u8; 2500],
-}
-
-#[inline]
-fn pos_as_idx(x: u8, y: u8) -> usize {
-    (x as usize) * 50 + (y as usize)
-}
-
-#[inline]
-fn idx_as_pos(idx: usize) -> (u8, u8) {
-    ((idx / 50) as u8, (idx % 50) as u8)
+    #[serde(with = "serde_impls")]
+    bits: [u8; ROOM_AREA],
 }
 
 impl Default for LocalCostMatrix {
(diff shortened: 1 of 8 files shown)


src/local/room_coordinate.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+#[inline]
+pub fn xy_to_linear_index(xy: RoomXY) -> usize {
+    ((xy.x.0 as usize) * (ROOM_SIZE as usize)) + (xy.y.0 as usize)
+}
+
+#[inline]
+pub fn linear_index_to_xy(idx: usize) -> RoomXY {
+    RoomXY {
```
khoover, line 22: Maybe I should check that idx is a valid linear index, then use the unchecked versions below? Avoids missing an overflow after the downcast.

src/local/room_coordinate.rs
```diff
@@ -0,0 +1,136 @@
...
+        x: unsafe { RoomCoordinate::unchecked_new((idx / (ROOM_SIZE as usize)) as u8) },
+        y: unsafe { RoomCoordinate::unchecked_new((idx % (ROOM_SIZE as usize)) as u8) }
+    }
+}
+
+#[derive(Debug, Hash, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct RoomCoordinate(u8);
+
+#[derive(Debug, Hash, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
```
khoover, line 36: I could write a custom serialize/deserialize that uses a single u16 for binary formats and a struct with derive impls for human readable formats? Unclear what kind of savings that gives, but if RoomXY ends up being something commonly stored on the wire it's better to have.

src/local/cost_matrix.rs
```diff
@@ -172,28 +162,24 @@ impl SparseCostMatrix {
         SparseCostMatrix { inner: HashMap::new() }
     }
 
-    pub fn get(&self, x: u8, y: u8) -> u8 {
-        assert!(x < 50, "out of bounds x: {}", x);
-        assert!(y < 50, "out of bounds y: {}", y);
-        if let Some(ref_val) = self.inner.get(&(x, y)) {
+    pub fn get(&self, xy: RoomXY) -> u8 {
+        if let Some(ref_val) = self.inner.get(&xy) {
```
Azaril, line 166: This can also be done as: ``
self.inner.get(&gy).as_deref().unwrap_or(0)
``

src/local/room_coordinate.rs
```diff
@@ -0,0 +1,136 @@
...
+    assert!(idx < ROOM_AREA, "Out of bounds index: {}", idx);
+    // SAFETY: bounds checking above ensures both are within range.
+    RoomXY {
+        x: unsafe { RoomCoordinate::unchecked_new((idx / (ROOM_SIZE as usize)) as u8) },
+        y: unsafe { RoomCoordinate::unchecked_new((idx % (ROOM_SIZE as usize)) as u8) }
+    }
+}
+
+#[derive(Debug, Hash, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct RoomCoordinate(u8);
```
Azaril, line 34: Likely want: ``#[serde(transparent)]``

src/local/room_coordinate.rs
Azaril: I suspect people will store this type a lot in memory if it becomes the defacto way of expressing an XY in valid room bounds within the API. (i.e. I know I'd convert over my code base.) I'd recommend adding custom serialization here.

src/local/room_position.rs
```diff
@@ -263,8 +260,8 @@ impl Position {
     pub fn from_packed(packed: i32) -> Self {
```
Azaril, line 260: Looking at the types, is there a reason this is an i32?

src/local/room_coordinate.rs
```diff
@@ -0,0 +1,136 @@
...
+    // # Safety
+    // Calling this method with `coord >= 50` can result in undefined behaviour when used.
+    #[inline]
+    pub unsafe fn unchecked_new(coord: u8) -> Self {
+        debug_assert!(coord < ROOM_SIZE, "Out of bounds unchecked coordinate: {}", coord);
+        RoomCoordinate(coord)
+    }
+
+    #[inline]
+    pub fn val(self) -> u8 {
```
Azaril, line 58: I would recommend implementing Deref or just using into() which is more standard. Looking at the rest of the code 'val()' becomes very repetitive - using more idiomatic rust make it more obvious. Additionally you could get rid of some of the other casts where the 'into' output type could be inferred.

Azaril (changes requested):

src/local/room_coordinate.rs
khoover: Don't we want to do the validation on deserializing too? That's why it has the custom impls for serialize and deserialize.

src/local/room_position.rs
khoover: Not as far as I know, @shanemadden would have more insight.

src/local/room_coordinate.rs
khoover: It has the From<RoomCoordinate> for u8, so sure, this becomes redundant.

src/local/cost_matrix.rs
khoover: I can't believe I missed that, makes some RefCell-like structs I was writing much cleaner too.

src/local/cost_matrix.rs
```diff
@@ -3,23 +3,16 @@ use std::convert::TryInto;
...
+use serde::{Serialize, Deserialize};
+
 use crate::objects::CostMatrix;
 
-use super::Position;
+use super::{Position, xy_to_linear_index, linear_index_to_xy, RoomCoordinate, RoomXY, ROOM_AREA};
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct LocalCostMatrix {
```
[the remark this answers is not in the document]
khoover: Serde doesn't do array serialization automatically, only slices, hence the serde_impl module.

src/local/room_position.rs
shanemadden: Huh! That's pulled straight forward from the current master branch unaltered, but you're right, that _shouldn't_ work. We have tests using negative numbers, but I guess they're not correct packed positions? Will dig a bit more and confirm.

src/local/room_coordinate.rs
Azaril: Good point - I think you get the same result given you just serialize the internal value.

Azaril (approved):

shanemadden (approved):

shanemadden:
Manually squash-merged into the bindgen branch due to the build failure not letting me do it here, and it didn't associate properly to close the PR 🤦 

Thanks for the changes, this will be a big improvement!

PR #368, opened 2021-12-02
Manually implemented RawObjectId de/serialization to/from string

agg23:
The standard `serde` de/serialize implementations stored the `RawObjectId` `u128` as a limited precision number, such as `1.2193937513084523e+38`. This custom implementation stores the value instead as a human readable string.

We could instead serialize this as a byte array for efficiency, but since the `serde_json` format is a lower efficiency mechanism anyway, I think defaulting to readability is better.


Dessix:
I would advise implementing human-readable versus compact formats as used in serde-rs/serde#1044 to allow human-readable while also using compact-encode during binary serialization.

Dessix (changes requested):
Marking as changes-requested; see prior comment about the human_readable flag on serializers.

PR #387, opened 2023-03-31
Re-organize exports of overcrowded objects module

shanemadden:


ASalvail:
If you give it another pass before the 0.11 release, we could include this PR in it.

shanemadden:
Definitely missed this comment when I was prepping that release - should have included it, whoops!

PR #415, opened 2023-06-21
Add checked version of `Position::from_world_coords` and checked add operations.

asquared31415:
This also implements `Position::from_world_coords` and several other methods in terms of `checked_from_world_coords`, and adds `#[track_caller]` so that the file and line number of the *caller* is used when the coords are out of bounds.

All checked operations can be expressed in terms of `Position::world_coords` and `Position::checked_from_world_coords`, but for convenience, checked versions of the `Add` impls are added.

With optimizations enabled, the compiler appears to be able to determine the valid range of inputs that produce `Ok` and `Err` and determine the value of the `Position` or error.  In fact, a check over the entirety of the input space gets compiled out to succeed instantly!  This unfortunately probably cannot be as useful in cases where positions and offsets are dynamically created, especially from JS, but it appears that the compiler can tell that the output of this function is in bounds anyway.

### DIFF
--- a/src/local/position/extra_math.rs
+++ b/src/local/position/extra_math.rs
@@ -3,7 +3,7 @@
 use std::ops::{Add, Sub};
 
 use super::Position;
-use crate::constants::Direction;
+use crate::{constants::Direction, local::position::WorldPositionOutOfBoundsError};
 
 impl Position {
     /// Returns a new position offset from this position by the specified x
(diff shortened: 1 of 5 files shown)


shanemadden (approved):

src/local/position/extra_math.rs
```diff
@@ -55,9 +55,49 @@ impl Position {
...
+    ///
+    /// For a panicking variant of this function, see [`Position::add`].
+    ///
+    /// See [`Position::from_world_coords`].
+    #[inline]
+    pub fn checked_add(self, rhs: (i32, i32)) -> Result<Position, WorldPositionOutOfBoundsError> {
+        let (x1, y1) = self.world_coords();
+        let (x2, y2) = rhs;
+
+        Position::checked_from_world_coords(x1 + x2, y1 + y2)
```
khoover, line 79: `x1.checked_add(x2).ok_or(...)?`, here and elsewhere we directly take `(i32, i32)` pairs.

PR #417, opened 2023-06-25
Replace `ReturnCode` with `Result<(), ErrorCode>`

### LINKED ISSUE #3: Use `Result<(), ??>` over `ReturnCode`
Making an issue here as a reminder.

It seems generally be more "rusty" to do this. The one disadvantage is having a two-byte return type size rather than one-byte, but that shouldn't be too large of a problem at all. We could even leave `ReturnCode` as a type alias for `Result<(), ??>`.

One question I'm not sure about is what to call the error portion. I'm thinking `ErrorCode` would work well, but just `Error` or `GameError` might also be good. I'm partially against `Screeps*` names because of duplication with `screeps::` crate name.

shanemadden:
Resolves #3 🎊 

We've discussed in discord replacing `ErrorCode` with per-result-set error types to allow for exhaustive matching, but saving that for a separate PR since this is already a big change

### DIFF
--- a/src/constants/small_enums.rs
+++ b/src/constants/small_enums.rs
@@ -1,79 +1,17 @@
 //! Various constants translated as small enums.
-use std::{
-    convert::{Infallible, TryFrom},
-    fmt,
-    str::FromStr,
-};
+use std::{convert::Infallible, fmt, str::FromStr};
 
 use enum_iterator::Sequence;
 use js_sys::JsString;
 use num_derive::FromPrimitive;
-use num_traits::FromPrimitive;
 use serde::{Deserialize, Serialize};
 use serde_repr::{Deserialize_repr, Serialize_repr};
 use wasm_bindgen::prelude::*;
 
-use crate::constants::find::{Exit, Find};
-
-// Bindgen does not correctly handle i8 negative return values. Use custom
-// return values.
-/// Translates return code constants.
-#[derive(
-    Debug, PartialEq, Eq, Clone, Copy, Hash, FromPrimitive, Deserialize_repr, Serialize_repr,
-)]
-#[repr(i8)]
-pub enum ReturnCode {
-    Ok = 0,
-    NotOwner = -1,
-    NoPath = -2,
-    NameExists = -3,
-    Busy = -4,
-    NotFound = -5,
-    NotEnough = -6,
-    InvalidTarget = -7,
-    Full = -8,
-    NotInRange = -9,
-    InvalidArgs = -10,
-    Tired = -11,
-    NoBodypart = -12,
-    RclNotEnough = -14,
-    GclNotEnough = -15,
-}
-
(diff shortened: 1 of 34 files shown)


src/constants/small_enums.rs
```diff
@@ -97,26 +35,49 @@ pub enum ErrorCode {
...
+            -7 => Err(ErrorCode::InvalidTarget),
+            -8 => Err(ErrorCode::Full),
+            -9 => Err(ErrorCode::NotInRange),
+            -10 => Err(ErrorCode::InvalidArgs),
+            -11 => Err(ErrorCode::Tired),
+            -12 => Err(ErrorCode::NoBodypart),
+            -14 => Err(ErrorCode::RclNotEnough),
+            -15 => Err(ErrorCode::GclNotEnough),
+            // SAFETY: Return codes must always be one of the values already covered
+            _ => unsafe { std::hint::unreachable_unchecked() },
```
xilexio, line 59: Why not regular `unreachable!()` to prevent UB? Note that the trait is pub and in some circumstances, such as putting errors to some structures first, it is possible to mix them up.

src/traits.rs
```diff
@@ -10,12 +10,24 @@ use wasm_bindgen::prelude::*;
...
     constants::*,
     enums::*,
-    js_collections::JsObjectId,
     local::{ObjectId, Position, RawObjectId, RoomName},
     objects::*,
     pathfinder::SingleRoomCostResult,
+    prelude::*,
 };
 
+pub trait FromReturnCode {
```
xilexio, line 19: Why `ReturnCode` now that it does not exist? There should be a comment explaining that this return code is actually the i8 returned by wasm_bindgen calls. Might be a good idea to put `type ReturnCode = i8;` somewhere and use it everywhere in API for clarity.

src/constants/small_enums.rs
shanemadden: Maybe a crate feature that's off by default that turns on 'crazy enough to trust the game's return values to be what we expect' mode? Because that makes sense - we don't want this to be insanely difficult to debug in the case where we do actually get a possible return value wrong.

src/traits.rs
shanemadden: I suppose in my 'naming' thinking I had the mind that even with no more `ReturnCode` type, the concept of a 'return code' is the game's, the set of i8s that map to the specific error values in constants. I dunno that it'd ever be anything other than i8 or if it'd be useful to bind that type?

xilexio: Probably JS API return code will eternally fit in i8, but calling these i8s by `ReturnCode` alias might make the code more readable and it will be easier to search for `ReturnCode` occurrences when refactoring than to search for i8s. In the end, this won't matter much to the average end user as they will use new `Result<..., ErrorCode>` instead.

PR #423, opened 2023-07-02
Add `RoomName::checked_add`

asquared31415:
Adds a checked variant of `RoomName + (i32, i32)` that can be used to do math on room names instead of panicking.

Similar reasoning as #415 but `RoomName` had no public equivalent that would let you implement this manually. (Should the to/from coord methods maybe be public too?)

### DIFF
--- a/src/local/room_name.rs
+++ b/src/local/room_name.rs
@@ -156,6 +156,19 @@ impl RoomName {
         self.packed
     }
 
+    /// Adds an `(x, y)` pair to this room's name.
+    ///
+    /// # Errors
+    /// Returns an error if the coordinates are outside of the valid room name
+    /// bounds.
+    ///
+    /// For a panicking variant of this function, see [`RoomName::add`].
+    pub fn checked_add(&self, offset: (i32, i32)) -> Result<RoomName, RoomNameParseError> {
+        let (x1, y1) = (self.x_coord(), self.y_coord());
+        let (x2, y2) = offset;
+        Self::from_coords(x1 + x2, y1 + y2)
+    }
+
     /// Converts this RoomName into an efficient, stack-based string.
     ///
     /// This is equivalent to [`ToString::to_string`], but involves no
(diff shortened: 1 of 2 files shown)


src/local/room_name.rs
```diff
@@ -156,6 +156,19 @@ impl RoomName {
...
+    ///
+    /// # Errors
+    /// Returns an error if the coordinates are outside of the valid room name
+    /// bounds.
+    ///
+    /// For a panicking variant of this function, see [`RoomName::add`].
+    pub fn checked_add(&self, offset: (i32, i32)) -> Result<RoomName, RoomNameParseError> {
+        let (x1, y1) = (self.x_coord(), self.y_coord());
+        let (x2, y2) = offset;
+        Self::from_coords(x1 + x2, y1 + y2)
```
khoover, line 169: If this is really checked, we should be using `x1.checked_add(x2).ok_or(RoomNameParseError)?`.

khoover: Actually, maybe a bit better in terms of reducing how many `jne`s are needed:
```rust
let (x, y) = x1.checked_add(x2)
     .zip(y1.checked_add(y2))
     .ok_or(RoomNameParseError::TooLarge(...))?;
```

asquared31415: Oh wow, completely forgot about overflow when making changes that would prevent overflow in a different way! :laughing:

asquared31415: Actually: since `self.AXIS_coord()` is in the range `-128..=127`, adding any `i32` to it cannot overflow around the `i32` space and then back in bounds.  Every time that the function, as currently written, outputs `Ok`, no overflow has occurred, even in intermediate calculations.

khoover: Right, but as written it panics if either of the coords pairs are close enough to `i32::MIN/MAX` that overflow in that add occurs. Pathological cases, but checked adds shouldn't panic ever.

shanemadden (approved):

PR #425, opened 2023-07-04
Add `sim` feature to enable sim room name override

shanemadden:
The special case for the `sim` room going into and out of positions/room names is always on, at the moment - though for most mature bots, the ability to run in the sim room is unimportant. Add the `sim` feature, which must be enabled to retain support for the special cases encountered there.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,6 +2,12 @@
 //!
 //! # Cargo Features
 //!
+//! ## `sim`
+//!
+//! Enables special-case handling of the unique room name present in the
+//! simulator - must be enabled to build code that is compatible with that
+//! environment
+//!
 //! ## `generate-pixel`
 //!
 //! Enables the function to generate pixels, which is only present on the
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,7 +6,8 @@
 //!
 //! Enables special-case handling of the unique room name present in the
 //! simulator - must be enabled to build code that is compatible with that
-//! environment
+//! environment. If this is enabled, the top-left valid room coordinate has the
+//! name `sim`, otherwise it's named `W127N127`.
 //!
 //! ## `generate-pixel`
 //!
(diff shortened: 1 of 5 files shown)


src/local/room_name.rs
```diff
@@ -61,20 +61,22 @@ impl fmt::Display for RoomName {
         let x_coord = self.x_coord();
         let y_coord = self.y_coord();
 
+        #[cfg(feature = "sim")]
         if self.packed == 0 {
             write!(f, "sim")?;
+            return Ok(());
+        }
```
asquared31415, line 68:
```suggestion
        if cfg!(feature = "sim") && self.packed == 0 {
            write!(f, "sim")?;
            return Ok(());
        }
```
As discussed elsewhere, this is free and improves checks.

src/local/room_name.rs
```diff
@@ -328,6 +330,7 @@ impl FromStr for RoomName {
 }
 
 fn parse_to_coords(s: &str) -> Result<(i32, i32), ()> {
+    #[cfg(feature = "sim")]
     if s == "sim" {
         return Ok((-HALF_WORLD_SIZE, -HALF_WORLD_SIZE));
     }
```
asquared31415, line 336: This could be changed similarly

src/lib.rs
```diff
@@ -2,6 +2,12 @@
 //!
 //! # Cargo Features
 //!
+//! ## `sim`
+//!
+//! Enables special-case handling of the unique room name present in the
+//! simulator - must be enabled to build code that is compatible with that
+//! environment
```
asquared31415, line 9: Again, a note on what exactly the behavior changes are would be nice.

Maybe the relevant method docs should also mention "if the `sim` feature is enabled..." too?

asquared31415 (commented):
Added some more feedback along the lines that we discussed externally, and I think the docs could be better too. But otherwise looks helpful that bots can avoid special cases.

asquared31415 (approved):

PR #432, opened 2023-07-11
Wrap raw store functions

### LINKED ISSUE #429: Creep::store functions panic if creep has no active Carry parts
When a creep has no acitve Carry parts, e.g. due to damage, calls to Store::get_capacity and Store::get_free_capacity panic instead of returning 0. A message similar to the following is produced:

```
caught exception: Error: expected a number argument
--
  | stack trace: Error: expected a number argument
     at _assertNum (screeps-rust:125:39)
     at screeps-rust:603:5
     at logError (screeps-rust:216:18)
     at module.exports.__wbg_getFreeCapacity_b255ba7c46af8c39 (screeps-rust:601:77)
     at screeps::objects::impls::store::Store::get_free_capacity::h29ae5514c3662d7f (wasm-function[3472]:68)     at ferrous_screeps::commodity_miner::run::{{closure}}::h74b1aed5d30c114b (wasm-function[143]:1191)
     at std::thread::local::LocalKey::try_with::hf5f8
```

Ninetailed:
The get_ functions on Store can return null if the resource type is not valid for that store. This can cause unexpected panics if, e.g. a creep tries to check its store when all of its Carry parts have been destroyed.

Renamed the raw functions with _internal, following the pattern used elsewhere, made them accept Option<i/u32>s, and had the public functions unwrap those options. 0 is used as a default value, which makes conceptual sense for all funcitons: a store that can't take a resource has no capacity for it and currently stores none.

Resolves #429

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ Unreleased
 - Rename `RoomObject::pos` to `RoomObject::js_pos` to avoid confusion with `HasPosition::pos`
   and remove the possibiliy for differing behavior based on whether the trait was imported
   (breaking)
+- Fix potential for panic in store functions when called with resource types that the store
+  isn't currently valid for
 
 0.14.0 (2023-07-03)
 ===================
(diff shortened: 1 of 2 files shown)


shanemadden (approved):
Thanks!!

PR #443, opened 2023-08-18
add compressed local raw terrain accessors with better types

asquared31415:
**This is not yet ready to be merged**

Adds a `RawRoomTerrain` to the `local` module, which wraps the `raw_terrain` screeps API call with a type that compresses the in-memory representation and uses the `RoomXY` and `Terrain` types.

API:
`fn get_raw(&RoomTerrain) -> RawRoomTerrain`
`fn get(&RawRoomTerrain, RoomXY) -> Terrain`


asquared31415:
The reduced memory overhead does not seem to be worth the extra processing cost. Closing in favor of #449

PR #449, opened 2023-09-13
Add native terrain data wrapper

### LINKED ISSUE #441: Add local representation of terrain objects like LocalCostMatrix
We should add a local struct type representing terrain that can be easily converted into, with the appropriate helper functions to handle the fact that the x/y coordinates are reversed compared to the layout of the bytes in a cost matrix.

shanemadden:
Add `LocalRoomTerrain`, a simplified alternative to the packed version in #443 (which we're planning to include in the utils crate for when a more compressed representation is desired) to hold the terrain natively in wasm memory for fast access. Resolves #441

### DIFF
--- a/src/local/terrain.rs
+++ b/src/local/terrain.rs
@@ -53,6 +53,7 @@ impl From<RoomTerrain> for LocalRoomTerrain {
         // again
         drop(js_buffer);
         // we've got the data in our boxed array, change to the needed type
-        LocalRoomTerrain::new_from_bits(unsafe { std::mem::transmute(data) })
+        // SAFETY: `Box` has the same layout for sized types. `MaybeUninit<u8>` has the same layout as `u8`. The arrays are the same size. The `MaybeUninit<u8>` are all initialized because JS wrote to them.
+        LocalRoomTerrain::new_from_bits(unsafe { std::mem::transmute::<Box<[MaybeUninit<u8>; ROOM_AREA]>, Box<[u8; ROOM_AREA]>>(data) })
     }
 }
--- a/src/local/terrain.rs
+++ b/src/local/terrain.rs
@@ -28,6 +28,9 @@ impl LocalRoomTerrain {
         }
     }
 
+    /// Creates a `LocalRoomTerrain` from the bytes that correspond to the room's terrain data.  This is like the `RoomTerrain` type but performs all operations on data stored in wasm.
+    /// Each byte in the array corresponds to the value of the `Terrain` at the given position.
+    /// The bytes are in row-major order, that is they start at the top left, then move to the top right, and then start at the left of the next row. This is different from `LocalCostMatrix`, which is column-major.
     pub fn new_from_bits(bits: Box<[u8; ROOM_AREA]>) -> Self {
         Self { bits }
     }
--- a/src/local/terrain.rs
+++ b/src/local/terrain.rs
@@ -6,6 +6,7 @@ use crate::{constants::Terrain, objects::RoomTerrain};
 
 use super::{xy_to_terrain_index, RoomXY, ROOM_AREA};
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct LocalRoomTerrain {
     bits: Box<[u8; ROOM_AREA]>,
 }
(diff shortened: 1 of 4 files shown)


src/local/terrain.rs
```diff
@@ -0,0 +1,58 @@
...
+        // create an uninitialized array of the correct size
+        let mut data: Box<[MaybeUninit<u8>; ROOM_AREA]> =
+            Box::new([MaybeUninit::uninit(); ROOM_AREA]);
+        // create a Uint8Array mapped to the same point in wasm linear memory as our
+        // uninitialized boxed array
+        let js_buffer =
+            unsafe { Uint8Array::view_mut_raw(data.as_mut_ptr() as *mut u8, ROOM_AREA) };
+        // SAFETY: it's important to not allocate _anything_ in rust memory now that
+        // we've created the Uint8Array pointed at an arbitrary wasm memory location;
+        // we'd overwrite the wrong memory if the array moved due to allocation!
```
asquared31415, line 45: As discussed in Discord, I believe that this safety requirement is incorrect (only a correctness requirement). However there is likely a safety requirement that the backing data isn't modified while the return value is used, much like `slice::from_raw_parts_mut`.  I have filed [an issue](https://github.com/rustwasm/wasm-bindgen/issues/3612) asking for clarification.

Also the safety comment (whatever its exact requirements and justification are) should be moved above the `unsafe` block.

src/local/terrain.rs
```diff
@@ -0,0 +1,58 @@
...
+        let byte = unsafe { self.bits.get_unchecked(xy_to_terrain_index(xy)) };
+        // not using Terrain::from_u8() because `0b11` value, wall+swamp, happens
+        // in commonly used server environments (notably the private server default
+        // map), and is special-cased in the engine code; we special-case it here
+        match byte {
+            0b00 => Terrain::Plain,
+            0b01 | 0b11 => Terrain::Wall,
+            0b10 => Terrain::Swamp,
+            _ => panic!("unexpected terrain"),
+        }
```
asquared31415, line 26: If we masked the byte with `& 0b11`, we could cause the panic to be optimized out. The only user facing change is that masking produces incorrect results if the terrain data ever updates (though we have so many other places that break anyway).
```suggestion
        match byte & 0b11 {
            0b00 => Terrain::Plain,
            0b01 | 0b11 => Terrain::Wall,
            0b10 => Terrain::Swamp,
            // Should be optimized out
            _ => unreachable!("all combinations of 2 bits are covered"),
        }
```

src/local/terrain.rs
```diff
@@ -0,0 +1,58 @@
+use std::mem::MaybeUninit;
+
+use js_sys::Uint8Array;
+
+use crate::{constants::Terrain, objects::RoomTerrain};
+
+use super::{xy_to_terrain_index, RoomXY, ROOM_AREA};
+
+pub struct LocalRoomTerrain {
```
asquared31415, line 9: This struct should at least be `Debug`, as is suggested for *all* structs.  Additionally, should it be `Clone` and (de)serializable?
```suggestion
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct LocalRoomTerrain {
```

src/local/terrain.rs
```diff
@@ -0,0 +1,58 @@
...
+        // map), and is special-cased in the engine code; we special-case it here
+        match byte {
+            0b00 => Terrain::Plain,
+            0b01 | 0b11 => Terrain::Wall,
+            0b10 => Terrain::Swamp,
+            _ => panic!("unexpected terrain"),
+        }
+    }
+
+    pub fn new_from_bits(bits: Box<[u8; ROOM_AREA]>) -> Self {
```
asquared31415, line 29:
```suggestion
    /// Creates a `LocalRoomTerrain` from the bytes that correspond to the room's terrain data.  This is like the `RoomTerrain` type but performs all operations on data stored in wasm.
    /// Each byte in the array corresponds to the value of the `Terrain` at the given position.
    /// The bytes are in row-major order, that is they start at the top left, then move to the top right, and then start at the left of the next row. This is different from `LocalCostMatrix`, which is column-major.
    pub fn new_from_bits(bits: Box<[u8; ROOM_AREA]>) -> Self {
```

asquared31415 (changes requested):
I have some concerns about the safety requirements of `view_mut_raw`, but I have requested clarification on the js_sys repo.

src/local/terrain.rs
shanemadden: I think the new safety comment should reflect the state of things as we understand it?

PR #454, opened 2023-09-18
Fixes that set visualize path style wasn't working

### LINKED ISSUE #453: Unable to set visualize path style
Hello. I am unable to set any visual path style with the rust api. It doesn't show any errors (or panics) but no visuals are displayed either.
I have the code ```creep.move_to_with_options(&source, Some(MoveToOptions::new().visualize_path_style(PolyStyle::default())))``` in my codebase but no visual path is shown. I also tried it with different PolyStyles. 

If I do the same thing in js it works fine (so it's not my client not showing them). Map visuals work through the rust api, as does letting the creeps say something.

Edit: Found the issue, looks like there is a typo for the setter here: https://github.com/rustyscreeps/screeps-game-api/blob/49b092c7c5167bfbf3463c401571cd464d759e55/src/objects/impls/creep_shared.rs#L24
It should be visualizePathStyle

I'll open up a PR to fix it

JomerDev:
Fixes a typo that prevented visualize_path_style from working. Closed #453

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased
 ==========
 
+- Fix incorrect setter name on `visualize_path_style` causing the setting to not work
+
 0.16.0 (2023-09-14)
 ===================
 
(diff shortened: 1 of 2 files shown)


shanemadden:
Whoops 😅  - thanks a bunch for tracking this down and for the PR!!

shanemadden (approved):

PR #455, opened 2023-09-21
feat: use gloo_utils for parsing

NekoFluff:
This PR does two things:

1. Use `JSValue::.into_serde()` from gloo utils. I was having problems with the `serde_wasm_bindgen::from_value`. More specifically, I kept seeing: `caught exception: SyntaxError: Unexpected token {`
2. By adding `Serialize` to the Path enum, we can easily call `move_by_path` using the serialized/stringified version of the Path.

e.g.
```rs
creep.move_by_path(&JsValue::from_str(&path.to_string()))
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ serde_json = "1"
 serde_repr = "0.1"
 wasm-bindgen = "0.2"
 serde-wasm-bindgen = "0.5"
+gloo-utils = "0.2.0"
 
 [dev-dependencies]
 assert_approx_eq = "1.1"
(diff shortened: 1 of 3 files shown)


NekoFluff:
Close it, merge it, w/e you wish. I couldn't get `serde_wasm_bindgen` working locally so I just wanted to leave this here if anyone else encountered a similar issue.

shanemadden:
Tried to reproduce this, and wasn't able to - can you provide any more info about how you're calling this that it's not working?  I'm doing a simple test of `find_path_to` from one `Position` to another with a default callback and it's deserializing the returned result happily via serde-wasm-bindgen.

PR #457, opened 2023-10-03
Fix `OwnedStructure::my` panicking on `undefined`

asquared31415:
This was (re-)discovered after the fix was lost in the transition to wasm bindgen.


asquared31415:
Confirmed to cause `my` on a freshly created controller to correctly return `false` under `profile="dev"` where it was throwing before.

shanemadden (approved):

PR #473, opened 2023-12-06
Documentation improvements

shanemadden:


asquared31415:
I suggest consistency in
`If an object in JavaScript memory is preferred, use <TYPE>`
versus
`Use <TYPE> if an object in JavaScript memory is preferred`

I don't particularly have an opinion on which one, just that it's consistent.

PR #479, opened 2024-01-02
fix: set_default_public_segment and set_public_segments are no longer panicking

pajotg:
This PR fixes 2 functions panicking on normal input
```
raw_memory::set_default_public_segment(None);
raw_memory::set_public_segments(&[0]);
```

I am not sure if my fix for `set_default_public_segment` is great
`None` needs to be translated to `null` on the JS side, previously it would be translated into `undefined` and would throw an error


shanemadden (approved):

shanemadden:
Thanks a bunch!!

PR #487, opened 2024-01-09
Replace forced `JsString` -> `String` conversions

shanemadden:
We've got a handful of places where the API assumes that the user is looking to pull a string across the boundary (which also requires conversion) - replacing those cases with `JsString` where appropriate, so that it's clear to the user that the conversion and boundary crossing is happening, and they can retain/use the string on the js side without bringing it across the boundary if that's more appropriate for their use case.

### DIFF
--- a/src/game/shard.rs
+++ b/src/game/shard.rs
@@ -9,27 +9,15 @@ extern "C" {
     #[wasm_bindgen(js_name = "shard")]
     type Shard;
 
+    /// Current shard name.
     #[wasm_bindgen(js_namespace = ["Game"], js_class = "shard", static_method_of = Shard, getter, js_name = name)]
-    fn name() -> JsString;
+    pub fn name() -> JsString;
 
+    /// Shard type. Currently always "normal".
     #[wasm_bindgen(js_namespace = ["Game"], js_class = "shard", static_method_of = Shard, getter, js_name = type)]
-    fn shard_type() -> JsString;
+    pub fn shard_type() -> JsString;
 
+    /// Flag for if this is a public test server or not.
     #[wasm_bindgen(js_namespace = ["Game"], js_class = "shard", static_method_of = Shard, getter, js_name = ptr)]
-    fn ptr() -> bool;
-}
-
-/// Current shard name.
-pub fn name() -> String {
-    Shard::name().into()
-}
-
-/// Shard type. Currently always "normal".
-pub fn shard_type() -> String {
-    Shard::shard_type().into()
-}
-
-/// Flag for if this is a public test server or not.
-pub fn ptr() -> bool {
-    Shard::ptr()
+    pub fn ptr() -> bool;
 }
--- a/src/game/shard.rs
+++ b/src/game/shard.rs
@@ -9,15 +9,27 @@ extern "C" {
     #[wasm_bindgen(js_name = "shard")]
     type Shard;
 
-    /// Current shard name.
     #[wasm_bindgen(js_namespace = ["Game"], js_class = "shard", static_method_of = Shard, getter, js_name = name)]
-    pub fn name() -> JsString;
+    fn name() -> JsString;
 
-    /// Shard type. Currently always "normal".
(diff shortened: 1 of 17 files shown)


src/lib.rs
```diff
@@ -128,7 +128,7 @@ pub use crate::{constants::*, enums::*, js_collections::*, local::*, objects::*,
 /// use js_sys::{JsString, Reflect};
 /// use screeps::{game, prelude::*, Creep};
 ///
-/// let c = game::creeps().get(String::from("Bob")).unwrap();
+/// let c = game::creeps().get("Bob".into()).unwrap();
```
asquared31415, line 131: `JsString::from` mirrors the `String::from` call that was here before, not that it matters, but it's interesting to change it.

src/game/shard.rs
```diff
@@ -9,27 +9,15 @@ extern "C" {
     #[wasm_bindgen(js_name = "shard")]
     type Shard;
 
+    /// Current shard name.
     #[wasm_bindgen(js_namespace = ["Game"], js_class = "shard", static_method_of = Shard, getter, js_name = name)]
-    fn name() -> JsString;
+    pub fn name() -> JsString;
```
asquared31415, line 14: In my opinion this has questionable utility, the most common use I can think of is `game::shard().name() == "shard3"` or similar, which does require the `String` version.

src/game/shard.rs
```diff
@@ -9,27 +9,15 @@ extern "C" {
...
 
+    /// Shard type. Currently always "normal".
```
asquared31415, line 16: I know in the docs it says that it's always `"normal"`, but is it worth specifying "in the MMO world..."

src/objects/impls/structure_portal.rs
```diff
@@ -62,7 +62,7 @@ extern "C" {
     fn room_internal(this: &InterShardPortalDestination) -> JsString;
 
     #[wasm_bindgen(method, getter)]
-    pub fn shard(this: &InterShardPortalDestination) -> String;
+    pub fn shard(this: &InterShardPortalDestination) -> JsString;
```
asquared31415, line 65: Similar thoughts here as `Shard::name`.

asquared31415 (commented):
I like the idea of using `JsString` more and I think that it will help especially in the case of creep names or segments.  I think that shard names are less useful as a `JsString` however, since most of the benefit from not converting is when you want to use a string as an "opaque key".  However I also see that it would be weird if the API was inconsistent, so transferring everything to `JsString` might make sense.

I don't have any *strong* objections either way, just some general thoughts.

src/raw_memory.rs
```diff
@@ -46,12 +46,21 @@ extern "C" {
...
 /// the previous tick, with segment numbers as keys and segment data in
 /// [`JsString`] form as values.
 ///
 /// [Screeps documentation](https://docs.screeps.com/api/#RawMemory.segments)
-pub fn segments() -> JsHashMap<u8, JsString> {
+pub fn segments() -> JsHashMap<u8, String> {
+    RawMemory::segments().into()
+}
+
+/// Get a [`JsHashMap<u8, String>`] with all of the segments requested on
```
asquared31415, line 58: wrong type in this doc comment

src/raw_memory.rs
```diff
@@ -46,12 +46,21 @@ extern "C" {
     fn set_public_segments(segment_ids: &Array);
 }
 
-/// Get a [`JsHashMap<u8, JsString>`] with all of the segments requested on
+/// Get a [`JsHashMap<u8, String>`] with all of the segments requested on
 /// the previous tick, with segment numbers as keys and segment data in
 /// [`JsString`] form as values.
```
asquared31415, line 51: this needs to be `String` in the doc comment

src/objects/impls/power_creep.rs
```diff
@@ -533,13 +543,20 @@ impl AccountPowerCreep {
         self.level_internal()
     }
 
-    /// Name of the power creep.
+    /// The power creep's name as an owned reference to a [`String`].
```
asquared31415, line 546: "owned reference" doesn't make sense, I would just say
> The power creep's name as a [`String`]

PR #490, opened 2024-01-12
Add missing docs link

asquared31415:


asquared31415:
The nightly clippy failure is a clippy bug, a new lint was introduced and there's many issues with it.  This case should be fixed by https://github.com/rust-lang/rust-clippy/pull/12137 when the next release comes out.

PR #499, opened 2024-02-29
add `RESOURCES_ALL` constant and `StoreObject::resource_types`

asquared31415:


asquared31415:
Nightly clippy is going to continue to fail because of the addition of `empty_doc_comment`, which apparently interacts poorly with `#[wasm_bindgen]`: https://github.com/rust-lang/rust-clippy/issues/12377

PR #500, opened 2024-02-29
Make ObjectId Send + Sync

scottbot95:
As suggested by @asquared31415 in discord, change the `PhantomData` on `ObjectId<T>` from `PhantomData<T>` to `PhantomData<fn() -T>`. This is more accurate since `ObectId` doesn't actually _own_ a `T`, but it can _produce_ a `T`.


scottbot95:
Nightly lint seems to be failing for the same reason as #499

shanemadden (approved):

PR #507, opened 2024-03-03
Remove generate-pixel and inter-shard-memory features, restrict more CPU functions

shanemadden:
CPU functions and data related to using CPU unlocks are missing on private servers, and attempting to use the interfaces will cause a panic - they should be behind a feature.

Instead of adding one or two more granular sub-features of `mmo`, it seemed better to just consolidate `generate-pixel` and `inter-shard-memory` down into the `mmo` feature. If a practical case arises where a private server supports some of the mmo-like interfaces but not all of 'em, it might make sense to re-add that level of granularity, but for now it seems like overkill.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Unreleased
 - Change name of `LocalRoomTerrain::get` to `get_xy` for consistency with `RoomTerrain`
 - Change name of `constants::extra::CONSTRUCTION_SITE_STOMP_RATIO` to
   `CONSTRUCTION_SITE_DROP_RATIO`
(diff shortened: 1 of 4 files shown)


scottbot95:
Might be worth also adding `#[cfg(not(feature="sim"))]` to the CPU halt functions `Game.cpu.halt()` does not exist in the sim.

tiennou:
> Might be worth also adding `#[cfg(not(feature="sim"))]` to the CPU halt functions `Game.cpu.halt()` does not exist in the sim.

There's a few more. I have `Game.map`, `Game.market` and `PathFinder.CostMatrix` also disabled in the profiler's default list there.

shanemadden:
Good point on the stuff that's not in sim - I'll make an issue to track getting those added as well!

PR #512, opened 2024-03-09
Refactor string enums to integer representations, add optional snippet optimizations

shanemadden:


shanemadden:
string enums have int reprs automatically in wasm-bindgen 0.2.93; no longer need some of this, but other parts will still be useful.

PR #521, opened 2024-06-16
(Improvement) Add additional calculation methods to RoomXY

jciskey:
This PR adds calculation methods to `RoomXY` that give it approximate parity to `Position`.

### DIFF
--- a/src/local/room_xy.rs
+++ b/src/local/room_xy.rs
@@ -1,9 +1,13 @@
-use std::fmt;
+use std::{cmp::Ordering, fmt};
 
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 
 use super::room_coordinate::{OutOfBoundsError, RoomCoordinate};
 use crate::constants::{Direction, ROOM_SIZE};
+
+mod approximate_offsets;
+mod extra_math;
+mod game_math;
 
 pub(crate) const ROOM_AREA: usize = (ROOM_SIZE as usize) * (ROOM_SIZE as usize);
 
@@ -67,6 +71,19 @@ pub struct RoomXY {
 }
 
 impl RoomXY {
+    /// Create a new `RoomXY` from a pair of `RoomCoordinate`.
+    #[inline]
+    pub fn new(x: RoomCoordinate, y: RoomCoordinate) -> Self {
+        RoomXY { x, y }
+    }
+
+    /// Create a new `RoomXY` from a pair of `u8`, checking that they're in
+    /// the range of valid values.
+    #[inline]
+    pub fn checked_new(x: u8, y: u8) -> Result<RoomXY, OutOfBoundsError> {
+        RoomXY::try_from((x, y))
+    }
+
     /// Create a `RoomXY` from a pair of `u8`, without checking whether it's in
     /// the range of valid values.
     ///
@@ -226,6 +243,19 @@ impl RoomXY {
         Direction::iter()
             .filter_map(|dir| self.checked_add_direction(*dir))
             .collect()
+    }
+}
+
+impl PartialOrd for RoomXY {
+    #[inline]
+    fn partial_cmp(&self, other: &RoomXY) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for RoomXY {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.y.cmp(&other.y).then_with(|| self.x.cmp(&other.x))
     }
 }
 
--- a/src/local/room_xy.rs
+++ b/src/local/room_xy.rs
@@ -255,7 +255,7 @@ impl PartialOrd for RoomXY {
 
 impl Ord for RoomXY {
     fn cmp(&self, other: &Self) -> Ordering {
-        self.y.cmp(&other.y).then_with(|| self.x.cmp(&other.x))
+        (self.y, self.x).cmp(&(other.y, other.x))
     }
 }
 
--- a/src/local/room_xy/approximate_offsets.rs
+++ b/src/local/room_xy/approximate_offsets.rs
@@ -1,0 +1,178 @@
+//! Methods related to approximating in-room positions
+//! between other in-room positions.
+
+use super::RoomXY;
+
+
+impl RoomXY {
+    /// Calculates an approximate midpoint between this point and the target.
+    ///
+    /// In case of a tie, rounds towards this point.
+    ///
+    /// If `distance_towards_target` is bigger than the distance to the target,
+    /// the target is returned.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use screeps::RoomXY;
+    ///
+    /// // Exact distances
+    /// let start = RoomXY::checked_new(10, 10).unwrap();
+    /// let target = RoomXY::checked_new(10, 15).unwrap();
+    /// assert_eq!(start.towards(target, 1), RoomXY::checked_new(10, 11).unwrap());
+    /// assert_eq!(start.towards(target, 4), RoomXY::checked_new(10, 14).unwrap());
+    /// assert_eq!(start.towards(target, 10), RoomXY::checked_new(10, 15).unwrap());
+    ///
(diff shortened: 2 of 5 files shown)


src/local/room_xy.rs
```diff
@@ -67,6 +71,19 @@ pub struct RoomXY {
 }
 
 impl RoomXY {
+    /// Create a new `RoomXY` from a pair of `RoomCoordinate`.
+    #[inline]
+    pub fn new(x: RoomCoordinate, y: RoomCoordinate) -> Self {
```
khoover, line 76: I suppose it's convention to have it, but it also feels weird given you can always do `RoomXY { x, y }` because of the pub fields.

src/local/room_xy.rs
```diff
@@ -229,6 +246,19 @@ impl RoomXY {
     }
 }
 
+impl PartialOrd for RoomXY {
```
khoover, line 249: Strongly disagree with these impls existing, considering even in-game objects disagree on being row or column major (which this implicitly enforces).

src/local/room_xy/approximate_offsets.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// assert_eq!(
+    ///     start.towards(target_1, 9),
+    ///     RoomXY::checked_new(14, 19).unwrap()
+    /// );
+    /// assert_eq!(
+    ///     start.towards(target_2, 1),
+    ///     RoomXY::checked_new(9, 10).unwrap()
+    /// );
+    /// ```
+    pub fn towards(self, target: RoomXY, distance_towards_target: i8) -> RoomXY {
```
khoover, line 52: This is a bit of a weird one; wouldn't you want a float in `[0,1]` for `self*x + target*(1-x)`?

src/local/room_xy/extra_math.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// assert_eq!(pos1 - pos2, (40, 20));
+    /// ```
+    #[inline]
+    fn sub(self, other: RoomXY) -> (i8, i8) {
+        let mx: i8 = self.x.0.try_into().unwrap();
+        let my: i8 = self.y.0.try_into().unwrap();
+        let ox: i8 = other.x.0.try_into().unwrap();
+        let oy: i8 = other.y.0.try_into().unwrap();
+
+        ((mx - ox), (my - oy))
```
khoover, line 165: Why not something like `(self.x.0.wrapping_sub(other.x.0) as i8, self.y.0.wrapping_sub(other.y.0) as i8)`?

src/local/room_xy.rs
jciskey: Honestly, future-proofing + convention following + "it feels really weird to build the struct in-place instead of using ::new"

src/local/room_xy.rs
jciskey: This was copied over from `Position`, purely because it allows for usage as a key in things like `BTreeMap` and that seems like a reasonable thing to keep parity with. If we want to tear it out here, we should tear it out there as well.

src/local/room_xy/approximate_offsets.rs
jciskey: I'm not entirely sure what you're asking here. Can you clarify a bit more?

src/local/room_xy/extra_math.rs
jciskey: Because it's not clear at first glance (or even second glance) that using `wrapping_sub` and then casting to `i8` is going to give you the proper negative offset. That didn't even come to mind, compared to just ensuring the arguments are all `i8` from the get-go and then subtracting them.

khoover: You're adding 4 extra comps and panic infrastructure to implement sub like this, though. And wrapping sub working like this is the entire reason signed arithmetic works between positive signed integers.

src/local/room_xy.rs
khoover: Honestly, yes. Position is even worse, since on top of picking row/column major, you also have to decide if you're doing a lex order on the (room name, room xy) pair or treating the whole position as a single world xy.

khoover (changes requested):
Left replies in discussions

src/local/room_xy.rs
khoover: For position, there's really 6 different ways you could arrange the components into a 4-tuple and use the lex ordering to come up with a sensible Position ordering:
- (room_name.x, room_xy.x, room_name.y, room_xy.y)
- (room_name.y, room_xy.y, room_name.x, room_xy.x)
- (room_name.x, room_name.y, room_xy.x, room_xy.y)
- (room_name.x, room_name.y, room_xy.y, room_xy.x)
- (room_name.y, room_name.x, room_xy.x, room_xy.y)
- (room_name.y, room_name.x, room_xy.y, room_xy.x)

src/local/room_xy.rs
```diff
@@ -229,6 +246,19 @@ impl RoomXY {
...
+    #[inline]
+    fn partial_cmp(&self, other: &RoomXY) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for RoomXY {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.y.cmp(&other.y).then_with(|| self.x.cmp(&other.x))
```
khoover, line 258: Nit; likely less efficient than
```suggestion
        (self.y, self.x).cmp(&(other.y, other.x))
```

src/local/room_xy/approximate_offsets.rs
jciskey: Ah, yes, this is a directional interpolation method. I'm not sure that returning the interpolation multiplier (or even vector) is all that helpful, though, since the user would still need to determine what `RoomXY` corresponds to that interpolation multiplier.

As with `PartialOrd`, this was copied from `Position` for API parity.

src/local/room_xy.rs
jciskey: Learn something new every day. I wasn't aware of the default implementation of short-circuiting comparisons for tuples. I agree that it's likely more efficient.

src/local/room_xy/extra_math.rs
jciskey: That's fair. For a core library like this, efficiency does matter more than code clarity.

src/local/room_xy.rs
jciskey: I can tear this out if we really want to, and you can open up a PR for tearing out the `Position` implementation, but I'm not comfortable doing the latter in this PR. It feels out of scope for the intended goal of bringing `RoomXY` into approximate API parity with `Position`.

shanemadden:
Back from a trip and catching up on this - the changes look good to me, though it sounds like whether to implement `Ord` and `PartialOrd` is a point of contention - or I should say *still* a point of contention, as the discussion around implementing it for `Position` and `RoomName` was contentious before my time in #226.  My inclination is to say that since those game types have `Ord` based on the outcome of that discussion then `RoomXY` might as well, until/unless we revisit that decision to have `Ord` on those?

jciskey:
That's pretty much where I'm sitting for the `Ord`/`PartialOrd` dispute. If `Position` has it, `RoomXY` should have it. If we want to ditch it entirely, that's a separate discussion.

khoover:
If we're going for consistency, then sure (although I'd also check that we used the same of column or row major here as we do for `Position`, away from an actual computer right now). I do think we'd be better served with something like `pub struct ColumnMajor(pub RoomXY/Position/RoomName); pub struct RowMajor...` instead of forcing a default choice.

jciskey:
> I'd also check that we used the same of column or row major here as we do for Position

It's the same ordering. `y` gets checked first, then `x`.

https://github.com/rustyscreeps/screeps-game-api/blob/main/src/local/position.rs#L376

shanemadden:
Got #523 opened for switching over to a struct-wrapped implementation for all of these ordered types when someone has a chance to implement that.

shanemadden (approved):

PR #526, opened 2024-08-06
Use `u8` for `RoomXY` ranges

shanemadden:


InfinityDevTech (approved):
This all looks good to me,

It made no sense to have it be a u32 to begin with, that distance could never be larger as the constructor for a RoomXY prevented it from ever being larger than 50.

The only thing I could ever see going wrong is if someone unsafely made a RoomXY over 50, but that is their fault if they used unchecked new.

PR #527, opened 2024-08-14
Stop using __Nonexhaustive for market type conversion

shanemadden:


Daholli:
Is there something I can help with here? Waanted to play around with it today, but stumbled upon this very error message

PR #532, opened 2024-08-18
Wrap the inner array for `LocalCostMatrix` with `Box`

shanemadden:


khoover:
I'm not sure if we want to do this; would remove the ability to do things like arena allocators or `Rc<LocalCostMatrix>` (or at least, add extra indirection to the latter).

PR #543, opened 2024-10-09
Improved `RoomCoordinate` and `RoomXY`, new `RoomOffset` for optimized arithmetic

khoover:
Adds indexing impls for `RoomCoordinate` and a wrapper type that allows indexing nested arrays by `RoomXY`. Re-writes `LocalCostMatrix` and `LocalRoomTerrain` indexing in terms of the `RoomCoordinate` indexing impls.

Also adds a new `RoomOffset` type which can be used for better-optimized arithmetic, as well as being the output of coordinate-coordinate subtraction.

### DIFF
--- a/src/local/room_coordinate.rs
+++ b/src/local/room_coordinate.rs
@@ -64,7 +64,7 @@ impl RoomCoordinate {
     /// Provides a hint to the compiler that the contained `u8` is smaller than
     /// `ROOM_SIZE`. Allows for better optimized safe code that uses this
     /// property.
-    pub fn assume_size_constraint(self) {
+    pub fn assume_bounds_constraint(self) {
         debug_assert!(self.0 < ROOM_SIZE);
         // SAFETY: It is only safe to construct `RoomCoordinate` when self.0 <
         // ROOM_SIZE.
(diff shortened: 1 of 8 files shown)


jciskey (approved):
LGTM, not seeing any major incompatibilities or issues. Even the documentation comments are solid.

khoover:
The broken nightly checks are in unrelated files, about converting explicit lifetimes into implicit ones.

khoover:
I think I can shave even more time off the distance transform in utils by using the `AsMut<XMajor<u8>> for LocalCostMatrix`.

src/local/room_coordinate.rs
khoover, line 1: There's actually some more stuff I'd like to add here, like `Deref<u8>` and a proper `RoomOffset` type for encoding coordinate-coordinate distances (would get to leverage it's restricted to `(-ROOM_SIZE, ROOM_SIZE)` when writing arithmetic).

khoover:
What I'd want to do is replace the `Sub` impl for `RoomXY` with one that returns an `OffsetXY` or similar, instead of `i8` pairs.

